PR #127, opened 2025-07-07
docs: Use generated schemas in documentation

bjchambers:
- Updated the rust code a bit to fix the generated schemas
- Fixed a config where `cargo test` was only testing `stepflow-main`
- Fixed test failures hidden by the above


docs/package.json
```diff
@@ -4,8 +4,9 @@
   "private": true,
   "scripts": {
     "docusaurus": "docusaurus",
-    "start": "docusaurus start",
-    "build": "docusaurus build",
+    "start": "npm run copy-schemas && docusaurus start",
+    "build": "npm run copy-schemas && docusaurus build",
+    "copy-schemas": "node scripts/copy-schemas.js",
```
bjchambers, line 9: Add a script to copy the schemas into the static site location.

schemas/protocol.json
```diff
@@ -2,12 +2,15 @@
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "title": "Message",
   "description": "The messages supported by the StepFlow protocol. These correspond to JSON-RPC 2.0 messages.\n\nNote that this defines a superset containing both client-sent and server-sent messages.",
-  "anyOf": [
+  "oneOf": [
```
bjchambers, line 5: Changed this to render a little better. Other updates were because of the missed tests allowing drift.

schemas/flow.json
```diff
@@ -56,8 +56,7 @@
     },
     "output": {
       "description": "The outputs of the flow, mapping output names to their values.",
-      "$ref": "#/$defs/ValueTemplate",
-      "default": null
+      "$ref": "#/$defs/ValueTemplate"
```
bjchambers, line 59: Missed test allowed drift. Fixed.

stepflow-rs/crates/stepflow-analysis/src/validation.rs
```diff
@@ -558,39 +558,6 @@ mod tests {
         );
     }
 
-    #[test]
```
bjchambers, line 561: Remove obsolete broken test. Missed due to issue.

stepflow-rs/Cargo.toml
```diff
@@ -9,7 +9,6 @@ repository = "https://github.com/riptano/stepflow"
 [workspace]
 resolver = "2"
 members = ["crates/*"]
-default-members = ["crates/stepflow-main"]
```
bjchambers, line 12: This "default-members" is what caused `cargo test` to only test `crates/stepflow-main`. I think we can safely remove it, and it will be better / avoid unwanted surprises.

stepflow-rs/crates/stepflow-core/src/workflow/flow.rs
```diff
@@ -257,10 +257,89 @@ impl JsonSchema for FlowRef {
         Flow::inline_schema()
     }
 }
+/// Configuration for testing a workflow.
```
bjchambers, line 260: Move this code in flow around so we have a single test block. I thought this was why the test wasn't running. I don't think it was, but still better to have a single test block.

PR #137, opened 2025-07-09
chore: add some tracing output at debug level for blob and value_reso…

zznate:
…lver, fix a setup issue with logging to allow passthrough of env variable

### DIFF
--- a/stepflow-rs/crates/stepflow-builtins/src/blob.rs
+++ b/stepflow-rs/crates/stepflow-builtins/src/blob.rs
@@ -71,12 +71,17 @@ impl BuiltinComponent for PutBlobComponent {
 
         let data_ref = ValueRef::new(input.data);
 
+        // DEBUG: Log what's being stored
+        tracing::debug!("put_blob storing data: {:?}", data_ref.as_ref());
+
         // Create the blob through the execution context
         let blob_id = context
             .state_store()
             .put_blob(data_ref)
             .await
             .change_context(BuiltinError::Internal)?;
+
+        tracing::debug!("put_blob created blob with ID: {}", blob_id.as_str());
 
         let output = PutBlobOutput {
             blob_id: blob_id.as_str().to_string(),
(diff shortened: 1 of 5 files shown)


stepflow-rs/crates/stepflow-builtins/src/blob.rs
```diff
@@ -71,13 +71,18 @@ impl BuiltinComponent for PutBlobComponent {
 
         let data_ref = ValueRef::new(input.data);
 
+        // DEBUG: Log what's being stored
+        tracing::debug!("put_blob storing data: {:?}", data_ref.as_ref());
```
bjchambers, line 75: I think `tracing::debug!("put_blob storing data: {data_ref:?}")` should work.

stepflow-rs/crates/stepflow-builtins/src/blob.rs
```diff
@@ -71,13 +71,18 @@ impl BuiltinComponent for PutBlobComponent {
...
+
         // Create the blob through the execution context
         let blob_id = context
             .state_store()
             .put_blob(data_ref)
             .await
             .change_context(BuiltinError::Internal)?;
 
+        tracing::debug!("put_blob created blob with ID: {}", blob_id.as_str());
```
bjchambers, line 84: Ditto -- I think the `blob_id` should wokr here (may need `{blob_id:?}` if we haven't implemented display, although we should support that if not).

stepflow-rs/crates/stepflow-core/src/values/value_resolver.rs
```diff
@@ -170,9 +172,12 @@ impl<L: ValueLoader> ValueResolver<L> {
         let path_result = if let Some(path) = expr.path() {
             match base_result {
                 FlowResult::Success { result } => {
+                    tracing::debug!("Resolving path '{}' on value: {:?}", path, result.as_ref());
```
bjchambers, line 175: Ditto -- generally prefer putting the references in the string -- `"Resolving path '{path}' on value: {result:?}"`.

stepflow-rs/crates/stepflow-core/src/values/value_resolver.rs
```diff
@@ -170,9 +172,12 @@ impl<L: ValueLoader> ValueResolver<L> {
...
                     if let Some(sub_value) = result.path(path) {
+                        tracing::debug!("Path '{}' resolved to: {:?}", path, sub_value.as_ref());
                         FlowResult::Success { result: sub_value }
                     } else {
+                        tracing::debug!("Path '{}' not found in value", path);
```
bjchambers, line 180: ditto

PR #138, opened 2025-07-09
feat: Support `.` and `[...]` in paths

bjchambers:
This allows a subset of JSON path expressions to be used as paths in StepFlow. We could extend this to support more of JSON path as needed.

### DIFF
--- a/stepflow-rs/crates/stepflow-core/src/workflow/expr.rs
+++ b/stepflow-rs/crates/stepflow-core/src/workflow/expr.rs
@@ -14,8 +14,7 @@
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-use crate::workflow::{json_path::JsonPath, ValueRef};
-
+use crate::workflow::{ValueRef, json_path::JsonPath};
 
 /// An expression that can be either a literal value or a template expression.
 #[derive(
@@ -319,7 +318,6 @@ mod tests {
             }
         );
     }
-
 
     #[test]
     fn test_expr_with_skip_action_from_yaml() {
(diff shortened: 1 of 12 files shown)


stepflow-rs/crates/stepflow-core/src/workflow/expr.rs
```diff
@@ -314,6 +320,7 @@ mod tests {
         );
     }
 
+
```
zznate: nit; uneccesary newline

stepflow-rs/crates/stepflow-core/src/workflow/json_path.rs
```diff
@@ -0,0 +1,347 @@
...
+        }
+    }
+
+    /// Parse a string path into a JsonPath
+    pub fn parse(s: &str) -> Result<Self, String> {
+        if s.is_empty() {
+            return Ok(Self::new());
+        }
+
+        let s = match s.trim().strip_prefix("$") {
```
zznate, line 71: Maybe pull out common strings into constants? Is rust better with optimising naked literals?

stepflow-rs/crates/stepflow-core/src/workflow/json_path.rs
```diff
@@ -0,0 +1,347 @@
...
+                    }
+
+                    if !found_end {
+                        return Err("Unclosed bracket '[' in path".to_string());
+                    }
+
+                    let bracket_content = bracket_content.trim();
+
+                    // Check if it's a quoted string
+                    if (bracket_content.starts_with('"') && bracket_content.ends_with('"'))
```
zznate, line 103: same here and the line below. Nits tho.

stepflow-rs/crates/stepflow-core/src/workflow/json_path.rs
bjchambers: Rust is pretty good about this -- they'll go into a single thing in the generated code. I think I'll keep these because I find it easier to read `"$"` and think "that is a dollar sign" than read `DOLLAR_SIGN`.

PR #149, opened 2025-07-16
feat: Add path transformations for routing

bjchambers:
This addresses two related problems:

1. Based on the route, the path to a component may be different. The component servers shouldn't need to be aware of routing to determine their components. The idea here is to allow the routing to update the path to reflect the path *relative the route* to pass to the component server. For instance, a route for `/python/*` can match `/python/bar` and pass `/bar` to the component server.
2. When we need to list all of the components available, we need to consider the routes. This means that if a route filters out some components, they should not be listed. If a route transforms the path to a component, it should be listed with the transformed path. If a route conditionally matches, that information should be displayed.

### DIFF
--- a/stepflow-rs/tests/routing/README.md
+++ b/stepflow-rs/tests/routing/README.md
@@ -1,0 +1,123 @@
+# Routing Tests
+
+This directory contains comprehensive tests and examples for the StepFlow routing system, demonstrating path transformation and component filtering capabilities.
+
+## Configuration Overview
+
+The `stepflow-config.yml` file demonstrates advanced routing features:
+
+### Complex Routing with Component Filtering
+```yaml
+- match: "/python/*"
+  target:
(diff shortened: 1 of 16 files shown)


stepflow-rs/tests/routing/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Configuration Overview
+
+The `stepflow-config.yml` file demonstrates advanced routing features:
+
+### Complex Routing with Component Filtering
+```yaml
+- match: "/python/*"
+  target:
+    plugin: python
```
bjchambers, line 13: This feels overly complicated. I'm trying to find a clean way to declare `/python/*` routes to the `python` plugin, and the component `/python/udf` corresponds to `/udf` within that plugin. The issue is how to describe that in a way that also allows printing the components, etc.

PR #168, opened 2025-07-25
feat: replace HTTP+SSE with Streamable HTTP

bjchambers:
This implements the newer Streamable HTTP transport described in the
MCP specification. The main benefit is that each request is sent via
POST and only use SSE for bidirectional communication when the component
uses `context` (indicating it may potentially call methods or send
notifications to the client).

Also adds the ability to start and stop test servers in the workflow
test configuration, allowing the HTTP tranpsort to be tested in
/tests/http.


stepflow-rs/crates/stepflow-protocol/src/plugin.rs
```diff
@@ -112,7 +110,7 @@ impl StepflowClientHandle {
                 .await
                 .change_context(PluginError::Execution),
             StepflowClientHandle::Http(client) => {
-                // Implement HTTP notifications using JSON-RPC notification format
+                // Streamable HTTP notifications using JSON-RPC notification format
```
bjchambers, line 113: ditto

stepflow-rs/crates/stepflow-protocol/src/http/client.rs
```diff
@@ -11,315 +11,537 @@
...
+            .owned_response()?;
+
+        // This is an assertion since the routing should only send the response for the
+        // registered ID to the pending one-shot channel.
+        debug_assert_eq!(response.response().id(), &id);
+        response.into_success_value()
+    }
+
+    /// Send a typed notification (fire-and-forget)
+    pub async fn notify<I>(&self, params: &I) -> Result<()>
```
bjchambers, line 127: Add a private `dyn` version with most of the code, so we don't need the many generic instantiations.

stepflow-rs/crates/stepflow-protocol/src/http/client.rs
```diff
@@ -11,315 +11,537 @@
...
+        expected_id: RequestId,
+        response: reqwest::Response,
+    ) -> Result<OwnedJson> {
+        let messages = self.sse_messages(response);
+        tokio::pin!(messages);
+
+        while let Some(owned_json) = messages.next().await {
+            match owned_json.message() {
+                Message::Request(request) => {
+                    // Handle bidirectional request from server in background
```
bjchambers, line 231: Spawn a task for this so we can handle multiple requests in parallel.

PR #175, opened 2025-07-28
feat: Added validate to CLI args. Supports both flow and config

zznate:


stepflow-rs/crates/stepflow-main/src/cli.rs
```diff
@@ -142,6 +143,27 @@ pub enum Command {
...
+            value_hint = clap::ValueHint::FilePath
+        )]
+        flow_path: Option<PathBuf>,
+
+        #[command(flatten)]
+        config_args: ConfigArgs,
+
+        /// Only validate the configuration file, not the workflow.
+        #[arg(long)]
+        config_only: bool,
```
bjchambers, line 161: Probably doesn't matter, but there is a way to make these mutually exclusive using the annotations (I don't remember it off the top of my head).

stepflow-rs/crates/stepflow-main/src/validate.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    if flow_only && flow_path.is_none() {
+        return Err(MainError::InvalidArgument(
+            "Must specify --flow when using --flow-only".to_string(),
+        )
+        .into());
+    }
+
+    // Validate configuration unless flow-only is specified
+    if !flow_only {
```
bjchambers, line 49: Can we validate a flow without a valid config? Eg., don't we need the config to know what paths are valid, etc.?

stepflow-rs/crates/stepflow-main/src/validate.rs
```diff
@@ -0,0 +1,288 @@
...
+    println!("  {} {}{}", level, diagnostic.text, path_str);
+}
+
+/// Print a formatted error message
+fn print_error(context: &str, error: &error_stack::Report<MainError>) {
+    println!("❌ {}: {}", context, error);
+}
+
+#[cfg(test)]
+mod tests {
```
bjchambers, line 242: Could also add some tests to the `command` tests (these test the CLI and use snapshots on the output).

stepflow-rs/crates/stepflow-main/src/validate.rs
zznate: Oh! Good point, i was thinking just syntactic validation, but like routes need to be established, named, etc.

zznate: I'll take out the optional arguments and just expect a stepflow-config.yml with each flow.

PR #214, opened 2025-08-14
docs: Examples update

zznate:


zznate:
missed a docs tag

examples/custom-python-components/workflow.yaml
```diff
@@ -41,7 +41,7 @@ steps:
       report_title: "Q4 Customer Analysis Report"
 
   - id: retrieve_report
-    component: "/builtin/get_blob"
+    component: "/get_blob"
```
bjchambers, line 44: Yeah, this has been a pain. I'm 50/50 on whether the pattern should be "builtins are registered as `/{*component}` and thus don't have a prefix and potentially collide with other things if the names are the same or "builtin plugin is registered at `/builtin/{*component}`.

We likely want to pick one as standard, and I suspect these are mix and match currently.

bjchambers (approved):

examples/custom-python-components/workflow.yaml
zznate: I dont think we need a prefix? The list isnt (and never really will be) exhaustive. If anything, should go the other way, like `/u/my_component` with a `/u` prefix (or whatever) denoting it's user defined.

PR #221, opened 2025-08-19
feat: Add a visualize command to the cli

bjchambers:
This uses graphviz and allows outputting DOT, SVG, PNG, etc.


stepflow-rs/crates/stepflow-main/src/visualize.rs
```diff
@@ -0,0 +1,440 @@
...
+    Svg,
+    /// PNG image format  
+    Png,
+}
+
+impl std::str::FromStr for OutputFormat {
+    type Err = String;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
```
zznate, line 40: nit: Maybe add trim on here for funzies/completeness?

zznate: Also, in doing a small deep dive on enums and why there is no valueOf equivalent (totally get it now), I learned about the `strum` crate which would allow something like the following:
```
use strum_macros::{Display, EnumString};

#[derive(Debug, Clone, Copy, PartialEq, Eq, Display, EnumString)]
#[strum(serialize_all = "lowercase")]
pub enum OutputFormat {
    Dot,
    Svg,
    Png,
}

// Now you get both for free:
let png = OutputFormat::Png;
println!("{}", png);  // prints: "png"
let parsed: OutputFormat = "png".parse().unwrap();  // works!
```

stepflow-rs/crates/stepflow-main/src/visualize.rs
```diff
@@ -0,0 +1,440 @@
...
+    pub async fn generate(&self, output_path: &Path) -> Result<()> {
+        let dot_content = self.generate_dot()?;
+        
+        match self.config.format {
+            OutputFormat::Dot => {
+                std::fs::write(output_path, dot_content)
+                    .change_context(MainError::Configuration)?;
+            }
+            OutputFormat::Svg => {
+                self.render_graphviz(&dot_content, output_path, "svg").await?;
```
zznate, line 101: Ie. per above, we could do:
```
    OutputFormat::Png => {
        self.render_graphviz(&dot_content, output_path, self.config.format.as_ref()).await?;
    }
```
But maybe im being pedantic.

stepflow-rs/crates/stepflow-main/src/visualize.rs
```diff
@@ -91,44 +79,43 @@ impl FlowVisualizer {
...
                 std::fs::write(output_path, dot_content)
                     .change_context(MainError::Configuration)?;
             }
-            OutputFormat::Svg => {
-                self.render_graphviz(&dot_content, output_path, "svg").await?;
-            }
-            OutputFormat::Png => {
-                self.render_graphviz(&dot_content, output_path, "png").await?;
+            format => {
+                self.render_graphviz(&dot_content, output_path, format)
```
zznate, line 89: Yay! That came out much cleaner!!

PR #349, opened 2025-10-08
feat: Kubernetes + load balancer demo

bjchambers:
- Use Lima VM to script deployment of a VM with k3s on OS X
- Add component server Docker image and Kubernetes manifests
- Configure Stepflow for k8s component routing
- Implement Pingora load balancer with instance-aware routing
- Add test script running simple components, bidirectional components and batch flows

**Pingora Load Balancer**
- Implement header-based instance routing
- SSE stream pass-through (no buffering)
- Backend discovery via health checks
- Least-connections load balancing
- Handle Stepflow-Instance-Id header
- Comprehensive design documentation

**Component Server Updates**
- Generate instance ID on startup (pod-name + uuid)
- Expose instance ID via health endpoint
- Add bidirectional component example (blob storage)
- Support StepflowContext for bidirectional communication

**Stepflow Runtime Updates**
- Propagate Stepflow-Instance-Id from SSE requests to responses

**Key Design Decisions**
- Use Stepflow-Instance-Id header (not URL-based routing)
- Instance = pod-name + startup-uuid
- Not resumable (new startup = new ID)
- Client tracks instance ID from SSE connection
- Load balancer routes responses back to same instance


examples/kubernetes-batch-demo/pingora-lb/src/main.rs
bjchambers, line 1: @zznate we may want to make this `pingora-lb` part of stepflow proper -- `stepflow-load-balancer` or something. It doesn't seem particularly specific to this example, and we're likely to want to release that binary, etc. as part of the Stepflow release process. Thoughts?

examples/kubernetes-batch-demo/docker/Dockerfile.component-server
```diff
@@ -0,0 +1,34 @@
...
+WORKDIR /app
+
+# Install uv for fast Python package management
+RUN pip install --no-cache-dir uv
+
+# Copy local stepflow-py SDK (build context is repo root)
+COPY sdks/python /tmp/stepflow-py
+
+# Install stepflow-py from local copy with HTTP support
+RUN uv pip install --system /tmp/stepflow-py[http]
```
bjchambers, line 15: This should change back to the released stepflow server once we have it updated.

examples/kubernetes-batch-demo/config/lima-k3s-qemu.yaml
```diff
@@ -0,0 +1,148 @@
...
+  mountPoint: "/home/lima.linux/stepflow"
+  writable: true
+
+# Network configuration (requires socket_vmnet)
+networks:
+- lima: shared
+
+# Provision script to install k3s and dependencies
+provision:
+- mode: system
```
bjchambers, line 33: To investigate: Can we put this in a `provision-lima-vm.yaml` and share it?

examples/kubernetes-batch-demo/config/lima-k3s-vz.yaml
```diff
@@ -0,0 +1,147 @@
...
+    echo "Stepflow project mounted at: /home/lima.linux/stepflow"
+
+# SSH configuration
+ssh:
+  localPort: 2222
+  loadDotSSHPubKeys: true
+
+# Port forwarding for k3s API and services
+# GRPC forwarding (recommended) - automatic port mapping
+portForwards:
```
bjchambers, line 135: I think we should be able to reduce this to just 80 -> 8080

examples/kubernetes-batch-demo/docker/Dockerfile.stepflow-server
```diff
@@ -0,0 +1,60 @@
+# Copyright 2025 DataStax Inc.
```
bjchambers, line 1: Should probably create a "proper" docker build file that we can use for stepflow (eg, multi-stage, cargo-chef, etc.)

examples/kubernetes-batch-demo/pingora-lb/src/main.rs
bjchambers: Created #351

examples/kubernetes-batch-demo/scripts/build-stepflow-server.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
bjchambers: #354

examples/kubernetes-batch-demo/docker/Dockerfile.component-server
bjchambers: #354

examples/kubernetes-batch-demo/docker/Dockerfile.stepflow-server
bjchambers: #352

PR #374, opened 2025-10-15
ci: Update release_rust for multiple images

bjchambers:
Cretae separate docker images for stepflow-server, stepflow-load-balancer and stepflow (the CLI).

### DIFF
--- a/stepflow-rs/release/Dockerfile.stepflow-load-balancer.alpine
+++ b/stepflow-rs/release/Dockerfile.stepflow-load-balancer.alpine
@@ -1,3 +1,17 @@
+# Copyright 2025 DataStax Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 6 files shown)


.github/workflows/release_rust.yml
```diff
@@ -162,22 +162,79 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - base: debian
+          # stepflow-server images
+          - image: stepflow-server
+            base: debian
```
zznate: Should we do alpine for musl (vs. glibc for debian?) - i guess these are different rust targets. I wonder which is most common, but I do know the alpine image will be smaller at least.

zznate: Oh lol, we are doing both, which seems to be the concensus, nevermind.

stepflow-rs/release/Dockerfile.stepflow-load-balancer.alpine
```diff
@@ -0,0 +1,24 @@
...
+# Create non-root user
+RUN adduser -D -s /bin/sh stepflow
+
+# Copy the stepflow-load-balancer binary (path provided by build context)
+COPY stepflow-load-balancer /usr/local/bin/stepflow-load-balancer
+RUN chmod +x /usr/local/bin/stepflow-load-balancer
+
+USER stepflow
+
+EXPOSE 8080
```
zznate: This is the default tomcat port. I know that traditionally, it was the address of NAT-ed HTTP back in the day, but we should def use something else.

Issue #367, opened 2025-10-09
Fix release docker files


bjchambers:
Currently, the entry point is `stepflow`. It should probably be `stepflow-server` or `stepflow-load-balancer`.

zznate:
`stepflow-server`, `stepflow-load-balancer` feels like the principal of least surprise to me and is real clear what's running - particularly when looking at a set of containers on a UI.

PR #377, opened 2025-10-15
ci: Update release script

bjchambers:
Specifically, try to fix issues with naming, and use a single (locally runnable) script for building images.


bjchambers:
This has been verified as fixing #367.

PR #393, opened 2025-10-21
feat: end-to-end distributed tracing for component servers

### LINKED ISSUE #271: Pass span-ID to component server for tracing purposes
This will allow component servers to emit tracing information that is properly tied back to the specific workflow step, etc.

- [ ] Pass trace ID, span ID, run ID, step ID to the component server
- [ ] Setup tracing and register these with tracing / logging as appropriate in Python SDK

bjchambers:
Add comprehensive distributed tracing support across the Rust runtime and component servers (Python SDK), enabling trace context propagation for bidirectional calls and full observability of workflow execution.

This enables full distributed tracing for debugging workflow execution, analyzing performance bottlenecks, and understanding cross-language execution flows in production environments.

This closes #271 by adding component-server tracing.

Screenshot 2025-10-21 at 9 46 28 AM

**Key Features:**
- Full trace correlation across Rust ↔ Python boundaries
- Automatic parent-child span relationships for bidirectional calls
- Run-level and step-level tracing granularity
- Zero-configuration tracing when STEPFLOW_OTLP_ENDPOINT is set
- Compatible with standard OTLP backends (Jaeger, Tempo, etc.)

**Protocol Extensions:**
- Add ObservabilityContext to component execution protocol
- Include trace_id, span_id, run_id, flow_id, and step_id
- Propagate context through component_execute and bidirectional calls
- Update protocol schema and Python SDK generated types

**Rust Runtime (stepflow-protocol):**
- Extract observability context from active spans using fastrace
- Inject context into ComponentExecuteRequest messages
- Propagate context through blob and flow handler bidirectional calls
- Add observability module with context extraction utilities

**Python SDK:**
- Add opentelemetry dependencies (tracing and logging)
- Implement ObservabilityConfig for OTLP configuration
- Auto-setup OTLP exporters from environment variables
- Propagate trace context to Python component execution spans
- Support both STDIO and HTTP server transports
- Add context injection for bidirectional calls back to runtime

**Integration Testing:**
- Add testcontainers for automatic OTLP collector management
- Implement isolated test environment with temp directories
- Auto-detect Colima Docker socket for macOS compatibility
- Add TRACING_TESTS.md documentation with troubleshooting
- Remove manual collector setup requirement

### DIFF
--- a/sdks/python/src/stepflow_py/observability.py
+++ b/sdks/python/src/stepflow_py/observability.py
@@ -48,6 +48,7 @@ logger = logging.getLogger(__name__)
 if typing.TYPE_CHECKING:
     from stepflow_py.generated_protocol import ObservabilityContext
 
+
 class ObservabilityConfig:
     """Configuration for OpenTelemetry observability.
 
@@ -182,7 +183,7 @@ def get_current_observability_context(
     run_id: str | None = None,
     flow_id: str | None = None,
     step_id: str | None = None,
-) -> "ObservabilityContext | None":  # type: ignore
+) -> "ObservabilityContext | None":
     """Capture the current OpenTelemetry span context for bidirectional requests.
 
     This extracts the trace_id and span_id from the current active span,
--- a/stepflow-rs/crates/stepflow-protocol/src/handlers/blob_handlers.rs
+++ b/stepflow-rs/crates/stepflow-protocol/src/handlers/blob_handlers.rs
@@ -15,10 +15,10 @@ use std::sync::Arc;
 use stepflow_plugin::Context;
 use tokio::sync::mpsc;
 
-use crate::error::{ TransportError};
-use crate::{Error, MethodHandler,};
-use crate::{ MethodRequest};
 use super::handle_method_call;
+use crate::MethodRequest;
+use crate::error::TransportError;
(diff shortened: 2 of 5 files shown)


stepflow-rs/crates/stepflow-server/tests/tracing_integration_tests.rs
```diff
@@ -0,0 +1,462 @@
...
+            println!(
+                "   File size: {} bytes",
+                std::fs::metadata(&trace_file).unwrap().len()
+            );
+        }
+        return;
+    }
+
+    // Look for workflow execution span
+    let workflow_span = find_span(&traces, "flow_execution");
```
bjchambers: Should also look for nested spans corresponding to the specific flow ID, and the bidirectional calls.

bjchambers: Even better would be to see if we could use the in-memory collector for these tests (although, that may be difficult since we need to connect to the execution layer).

Either way, we should modify these to:

a) Capture the run_id from the execution request, and use that to search for traces.
b) Capture the trace and sub-traces
c) Then look for more specific spans (including bidirectional calls)

This should help make sure we properly wire all the things.

PR #405, opened 2025-10-28
feat: 404 dashboard updates and add initial set of metrics, support for such in crates and config

zznate:


stepflow-rs/crates/stepflow-observability/src/lib.rs
```diff
@@ -126,6 +133,14 @@ impl ObservabilityConfig {
             );
         }
 
+        // Validate metrics_enabled requires otlp_endpoint
+        if self.metrics_enabled && self.otlp_endpoint.is_none() {
```
bjchambers: Probably worth updating the PR description -- it seems like the dashboard updates only affect our default observability stack -- the main feature is added support for metrics?

zznate: True - it's only like 2 metrics, but metrics none the less.

stepflow-rs/crates/stepflow-cli/tests/commands/snapshots/test_cli__commands__test_visualize__visualize_png_format.snap
```diff
@@ -1,18 +1,26 @@
...
+    STEPFLOW_TRACE_ENABLED: ""
 ---
-success: true
-exit_code: 0
+success: false
+exit_code: 1
 ----- stdout -----
 
 ----- stderr -----
-✅ Visualization generated: [TEMP_DIR]/stepflow_test_basic_output.png
```
bjchambers: Install graphviz.

PR #408, opened 2025-10-30
feat: http support for langflow integration

zznate:
Also increase pipe buffer sizes where possible


integrations/langflow/pyproject.toml
```diff
@@ -52,6 +52,13 @@ dependencies = [
     "nest-asyncio>=1.6.0",
 ]
 
+[project.optional-dependencies]
+http = [
```
bjchambers: These may not be necessary since Stepflow SDK has them? Or, maybe we need an optional dependency that enables an optional dependency of the stepflow SDK?

PR #414, opened 2025-11-05
feat: add mode-aware skip conditions for vector store workflows

bjchambers:
Implement sophisticated mode-aware execution for vector store workflows that allows fine-grained control over which steps execute in different modes (ingest, retrieve, hybrid). This enables efficient workflow execution by skipping unnecessary operations based on the execution context.

Key technical features:
- Mode-aware skip conditions using computational components (mode_check, mode_output) instead of expression evaluation
- Graph-based node classification algorithm that analyzes dependency relationships to determine if steps are part of ingestion or retrieval paths
- Propagation of classification from classified nodes to unclassified dependencies (e.g., embeddings feeding into vector stores)
- Graceful handling of skipped steps using OnSkipDefault for dependent steps
- MustExecute flag combined with default values for always-running output steps

Implementation details:
- Added mode_check component that returns boolean skip flags (skip_ingest_step, skip_retrieve_step) based on the execution mode
- Added mode_output component that provides appropriate feedback messages when retrieval is skipped in ingest-only mode
- Created _classify_nodes_for_mode_skipping() method in translator.py that uses graph traversal to identify ingestion vs retrieval steps
- Updated node_processor.py to apply skip conditions based on classification
- Enhanced vector store execution to support mode-aware output selection

Design decisions:
- Use computational approach (mode_check component) rather than expression evaluation since Stepflow's skipIf only accepts references to boolean values
- Iterate over all nodes during classification propagation, not just nodes with dependencies, to handle components like embeddings with no incoming edges
- Apply OnSkipDefault to retrieval_result input of mode_output step to allow execution even when retrieval is skipped
- Set mustExecute=True on all Langflow-converted steps and mode_output to ensure proper execution with skip conditions

### DIFF
--- a/integrations/langflow/README.md
+++ b/integrations/langflow/README.md
@@ -350,6 +350,64 @@ uv run stepflow-langflow execute tests/fixtures/langflow/vector_store_rag.json \
   }'
 ```
 
+### Vector Store Mode-Aware Execution
+
+Workflows with vector store components automatically support mode-aware execution, allowing you to control whether documents are ingested, retrieved, or both:
+
+#### Available Modes
+
+- **`ingest`**: Populate vector stores with documents (no retrieval)
+- **`retrieve`**: Search existing vector stores (assumes documents already ingested)
+- **`hybrid`** (default): Auto-detects based on inputs:
+  - If only documents provided → ingest
+  - If only query provided → retrieve
+  - If both provided → ingest then retrieve
+
+#### Mode Usage Examples
+
+```bash
+# Ingest mode: Load documents into vector store
+uv run stepflow-langflow execute tests/fixtures/langflow/vector_store_rag.json \
+  '{"mode": "ingest", "file_path": "documents.pdf"}' \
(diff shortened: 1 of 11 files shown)


integrations/langflow/README.md
```diff
@@ -350,6 +350,64 @@ uv run stepflow-langflow execute tests/fixtures/langflow/vector_store_rag.json \
...
+# Hybrid auto-detection: Omit mode parameter
+uv run stepflow-langflow execute tests/fixtures/langflow/vector_store_rag.json \
+  '{"message": "What topics are covered?", "file_path": "content.pdf"}'
+```
+
+#### When to Use Each Mode
+
+- **Ingest**: Bulk document loading, indexing pipelines, data preparation
+- **Retrieve**: Production queries, API endpoints, chatbot interactions
+- **Hybrid**: Interactive sessions, testing, exploratory workflows
```
zznate: Maybe explicitly call out chat and agentic use cases?

PR #416, opened 2025-11-08
feat: added options to sf-lf http server

zznate:


integrations/langflow/src/stepflow_langflow_integration/executor/langflow_server.py
```diff
@@ -17,13 +17,22 @@
...
 
+# Configure logging
+logging.basicConfig(
+    level=logging.INFO,
+    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+    force=True  # Override any existing configuration
+)
+
+
 class StepflowLangflowServer:
```
bjchambers: It feels like we shouldn't need this class given the pattern elsewhere, but probably OK?

integrations/langflow/src/stepflow_langflow_integration/executor/langflow_server.py
```diff
@@ -63,14 +72,36 @@ def run(self) -> None:
...
             port: Server port
+            workers: Number of worker processes
+            backlog: Maximum number of pending connections
+            timeout_keep_alive: Keep-alive timeout in seconds
         """
-        await self.server.start_http(host=host, port=port)
+        # Apply nest_asyncio to allow nested event loops in HTTP mode
+        # This is needed because Langflow components may call asyncio.run()
+        # from within an already-running event loop
+        import nest_asyncio
```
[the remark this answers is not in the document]
zznate: I *think* it has to be called every time before doing an event loop thing? Let me commit this and i'll take one or the other out once I know everything else works.

integrations/langflow/src/stepflow_langflow_integration/executor/langflow_server.py
zznate: Well, it encapsulates StepflowServer and UDFExecutor spin up with the component registration (and the StepflowContext through register_componets). I agree this is a little squirrely and we need to straighten this out between integrations and python sdk.

Issue #320, opened 2025-09-08
Add "globals" available within flows


bjchambers:
Within a flow, these would function like environment variables but could be scoped to specific users or set when running a flow. Ideally, they could be used to set default values for flow inputs, allowing (for instance) a per-user default value that is easily overridable.

Notes:
1. This may include the ability to initialize a set of global variables from the environment.
2. This could also include the ability to treat globals as secrets.
3. This may include the ability to use global variables within component servers (exposing them via the context). Specifically, this would allow cases like pulling the OPENAI_API_KEY from the global variable associated with the flow/user rather than an environment variable shared across all flows. If we do consider that, it could lead to some errors if we used the same component server for different values, so this may relate to routing and similar considerations. 

This would be useful in supporting global variables from Langflow, but we should decide whether it makes sense in Stepflow or is something to work around in the integration. And if we do add it, we should consider what form it takes, especially related to 3 above.

bjchambers:
> Agree that override, cleanup, and isolation are going to be big issues for common stuff like `OPEN_API_KEY`. We are going to need a fancy way of name spacing to provide isolation and security; simple prefixing is not going to cut it as those will be discoverable in any re-used/shared runtime.

Not sure I follow. I think we had discussed the globals being associated with a specific scope -- so we don't need to prefix these as much as just ensure that you can only access globals from your scopes. The bigger issue is preventing those being used to create some object that *is* shared (eg., preventing re-use). I *think* maybe we're saying the same thing with different words, but wanted to make sure.

bjchambers:
Refining based on conversations with @jordanrfrazier.

## Stepflow

#421 implements variables and secrets for stepflow. specifically:
1. Adds variables (with a schema) to the flow and the ability to reference them.
2. Adds the ability to mark variables as secrets and have them redacted (#422 looks at extending this to flow/step inputs/outputs).

## Langflow

1. If input field `_input_type is SecretStrInput` or `password: true`, then mark the field as a secret (so it is redacted and doesn't show up in logs).
2. If input field has `load_from_db: true` then reference a variable and use the value as the name.
3. Collect the referenced variables then to configure the flow variables. I can type them as `T | None` so they are optional. Make sure they agree on the types.
4. Have the langflow UDF fall back to the environment variable if the referenced variable isn't set (eg., if the input is `None`).

PR #424, opened 2025-11-19
feat: use variables for load_from_db inputs

bjchambers:
This uses Stepflow variables to implement `load_from_db` inputs.
This allows passing the variable when executing the flow.

This is part of #320 (specifically, part 2, 3 and 4) of the langflow
steps identified https://github.com/stepflow-ai/stepflow/issues/320#issuecomment-3549510640).

This also simplifies some of the python code (removing dead code) and fixes a bug in JsonPath (where `null` wasn't parsing as an empty JsonPath).

### DIFF
--- a/.github/actions/langflow-checks/action.yml
+++ b/.github/actions/langflow-checks/action.yml
@@ -36,8 +36,7 @@ runs:
     - name: Configure Rust cache
       uses: Swatinem/rust-cache@v2
       with:
-        workspaces: |
-          stepflow-rs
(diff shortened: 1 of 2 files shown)


bjchambers:
> Extraneous print statement. Also, seems like there was a lot of functionality removed from node_processor.py that I dont see picked up anywhere else.

The extraneous print is in tests, where logging is usually not configured. I think it is ok / expected it remain a print.

The functionality removed was not being used anywhere, so I deleted it as simplification.

Issue #446, opened 2025-12-15
Rework value expressions and execution


bjchambers:
# Update Value Templates

Rather than having a JSON value that contains references, there are benefits to having a top-level expression that can contain JSON. Specifically, this will make the JSON schema describe much better the format of the references, and it will make it easier to extend with new kinds of expressions.

* Old: `{$from: { step: <id> }}` becomes `{$step: <id>}`
* Old: `{$from: { step: <id> }, path = <path>}` becomes `{$step: <id>, path = <path>}`
* Old: `{$from: {workflow: input, path = "foo"}}` becomes: `{$workflow: "foo"}`

# New Edge Expressions

* `{$if: <expr>, then: <expr>, else: <expr> }` for conditional execution.
* `{$coalesce: [<expr>, <expr>, ...]}` for selecting the first successful, non-null result.

# Lazy Evaluation
Only expressions which are needed are evaluated. We already *mostly* have this, but this will lead to us getting rid of `skipIf` and the entire notion of "skipped" and instead replace it with conditional execution that makes the step unnecessary.

PR #451, opened 2025-12-16
feat!: Introduce new edge syntax

### LINKED ISSUE #447: Introduce value expressions containing literals
Currently, we allow the `{ $from: ... }` to appear anywhere in a *value template*. This makes it difficult to explain the dependencies of a step and doesn't handle the most common case -- outputs of one step flowing to the inputs of the next step. This sets the stage for conditions (`$if`, `$coalesce`, etc.) appearing as part of these expressions.

bjchambers:
- Introduces the new `{$step: <id>}` syntax (replacing `{$from: {step: <id>}}` and other syntax as described in #446.
- Introduces initial implementations of `$if` and `$coalesce`
- Updates execution and tests to use the new forms
- Updates all workflows to the new forms
- Updates schema to document the new expression
- Updates Python SDK and builders to use the new expressions
- Updates langflow integration to use the new forms
- Updates docs to use the new forms
- Updates visualization to support the new forms and render `$if` and `$coalesce` with diamonds

This closes #447.

### DIFF
--- a/integrations/langflow/uv.lock
+++ b/integrations/langflow/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.11"
 resolution-markers = [
     "python_full_version >= '3.13'",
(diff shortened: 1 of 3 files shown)


sdks/python/src/stepflow_py/__init__.py
```diff
@@ -57,16 +57,16 @@
     "Step",
     "Component",
     "Schema",
-    "ValueTemplate",
-    # Error and Skip Action types
+    # ValueExpr builder (new syntax with $step, $input, $variable)
```
bjchambers: Drop comment

stepflow-rs/crates/stepflow-cli/src/visualize.rs
```diff
@@ -257,38 +259,171 @@ impl FlowVisualizer {
         Ok(())
     }
 
+    /// Generate a unique ID for an expression node
+    fn next_expr_id(&self) -> String {
+        let id = self.expr_node_counter.get();
```
bjchambers: Should use atomic counter and/or `&mut self`.

PR #461, opened 2025-12-23
feat: Add static type checking for workflows

### LINKED ISSUE #438: Validation should verify and propagate schemas
This will require manipulating and propagating schemas. Ideally, the user could omit schemas and a minimal input schema (meeting the requirements) would be generated, and same for the output schema. Then, I think we could just confirm that the user-specified schema (if there is one) is compatible with the inferred schemas.

This will likely require interacting with the plugins/component servers to get the input/output types of a component. This will be especially important when working with "higher-order" components, such as executing a flow. If the flow is a literal, we should be able to use the input and output type of the flow to determine the input and output of the execution (likely passing this to the plugin and/or component servers). If the flow is *not* a literal, we may require some user specified type annotations at those points for now.

bjchambers:
Implements a type checking system for Stepflow workflows that can infer and validate types statically without executing flows.

## New crate: stepflow-typecheck

- Type representation with Schema, Any, and Never types
- Type environment tracking input, variable, and step output schemas
- JSON Schema structural subtyping (width, depth, array covariance)
- Expression type synthesis for all ValueExpr variants
- Flow-level type checking with forward propagation
- Input schema inference from workflow $input references

## CLI: stepflow infer

New command that analyzes a workflow and outputs an annotated version with inferred schemas:

    stepflow infer --flow=workflow.yaml
    stepflow infer --flow=workflow.yaml --output  # writes .inferred.yaml
    stepflow infer --flow=workflow.yaml --strict  # errors on untyped outputs

## Protocol: components/infer_schema

New protocol method allowing components to report output schemas based on their input types. Python SDK and builtins implement this for schema-aware type inference.

## Schema consolidation

FlowSchema now serializes as a valid JSON Schema with type definitions under $defs and flow-specific schemas (input, output, variables, steps) under properties. Uses custom Serialize/Deserialize for proper format.

Closes #438

### DIFF
--- a/docs/docs/best-practices/design.md
+++ b/docs/docs/best-practices/design.md
@@ -267,43 +267,45 @@ Use comprehensive input and output schemas:
 ```yaml
 name: "User Data Processor"
 
-input_schema:
+schemas:
   type: object
   properties:
-    user_id:
-      type: string
-      pattern: "^[a-zA-Z0-9]{8,}$"
-      description: "Unique user identifier"
-    processing_options:
+    input:
       type: object
       properties:
-        include_analytics:
-          type: boolean
-          default: true
-        output_format:
+        user_id:
           type: string
-          enum: ["json", "xml", "csv"]
-          default: "json"
+          pattern: "^[a-zA-Z0-9]{8,}$"
+          description: "Unique user identifier"
+        processing_options:
+          type: object
+          properties:
+            include_analytics:
+              type: boolean
+              default: true
+            output_format:
+              type: string
+              enum: ["json", "xml", "csv"]
+              default: "json"
+          additionalProperties: false
+      required: ["user_id"]
       additionalProperties: false
-  required: ["user_id"]
-  additionalProperties: false
-
-output_schema:
-  type: object
-  properties:
-    processed_data:
-      type: object
-      description: "Processed user data"
-    metadata:
+    output:
       type: object
       properties:
-        processing_time_ms:
-          type: integer
-          minimum: 0
-        version:
-          type: string
-      required: ["processing_time_ms", "version"]
-  required: ["processed_data", "metadata"]
+        processed_data:
+          type: object
+          description: "Processed user data"
+        metadata:
+          type: object
+          properties:
+            processing_time_ms:
+              type: integer
+              minimum: 0
+            version:
+              type: string
+          required: ["processing_time_ms", "version"]
+      required: ["processed_data", "metadata"]
 ```
 
 ### Validate at Step Level
--- a/docs/docs/components/component-server/custom-components.md
+++ b/docs/docs/components/component-server/custom-components.md
@@ -92,11 +92,15 @@ Reference your components in workflow YAML files:
 
 ```yaml
 schema: https://stepflow.org/schemas/v1/flow.json
-input_schema:
+
+schemas:
   type: object
   properties:
-    user_name:
-      type: string
+    input:
+      type: object
+      properties:
+        user_name:
+          type: string
 
 steps:
 - id: greeting_step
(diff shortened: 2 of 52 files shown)


sdks/python/src/stepflow_py/flow_builder.py
```diff
@@ -385,14 +375,27 @@ def build(self) -> Flow:
...
 
+        # Build FlowSchema if any schemas are set
+        schemas = None
+        if (
+            self.input_schema is not None
+            or self.output_schema is not None
+            or self.variables_schema is not None
+        ):
+            schemas = FlowSchema(
+                defs={},
```
bjchambers: We may want to allow adding defs to the schema (eg., for things like ChatMessage which may be referenced in multiple steps). We may also want to support step schemas. Perhaps something like an optional `schema` argument when adding steps? And a dict of defs which can then be added to the defs list if new?

stepflow-rs/crates/stepflow-cli/src/validate.rs
```diff
@@ -10,34 +10,58 @@
...
+        ConfigArgs { config_path: None }.load_config(flow_dir)?
     };
 
     // Use unified validation
-    let diagnostics = validate_with_config(&flow, &config.plugins, &config.routing)
+    let mut diagnostics = validate_with_config(&flow, &config.plugins, &config.routing)
         .change_context(MainError::ValidationError("Validation failed".to_string()))?;
 
+    // Optionally run type checking
+    if type_check {
```
bjchambers: Open an issue to create a more centralized cache for schema information (at least in the service). Perhaps validate should call a `validate` endpoint in the server?

stepflow-rs/crates/stepflow-cli/tests/commands/snapshots/test_cli__commands__test_infer__infer_secret_propagation.snap
```diff
@@ -0,0 +1,65 @@
...
+    STEPFLOW_OTLP_ENDPOINT: ""
+    STEPFLOW_TRACE_ENABLED: ""
+---
+success: false
+exit_code: 1
+----- stdout -----
+📊 Validation results: 0 fatal, 2 errors, 0 warnings
+
+❌ Errors:
+  ERROR Undefined variable 'api_key' referenced in flow ()
```
bjchambers: Declare these variables so we don't get the error.

stepflow-rs/crates/stepflow-cli/tests/commands/snapshots/test_cli__commands__test_infer__infer_secret_propagation.snap
```diff
@@ -0,0 +1,65 @@
...
+    path: $.messages
+
+
+----- stderr -----
+
+📥 Synthesized step input types:
+  create_message: { user_prompt: string, metadata: { api_key: any, temperature: any } }
+
+📋 Component expected input types:
+  create_message: any
```
bjchambers: Should fix the display of inferred types here, and also indicate whether things are secrets?

stepflow-rs/crates/stepflow-typecheck/src/error.rs
```diff
@@ -0,0 +1,256 @@
...
+    #[error("component '{component}' does not provide output schema")]
+    UntypedOutput {
+        /// The component that lacks output schema.
+        component: String,
+    },
+}
+
+impl TypeError {
+    /// Create a type mismatch error.
+    pub fn mismatch(expected: Type, actual: Type, detail: impl Into<String>) -> Self {
```
bjchambers: Probably easier to create directly to benefit from named parameters (for this and other type errors).

stepflow-rs/crates/stepflow-typecheck/src/input_inference.rs
```diff
@@ -0,0 +1,934 @@
...
+                    source_step,
+                );
+            }
+        }
+    }
+}
+
+/// A trie node for building nested schemas from path constraints.
+#[derive(Debug, Default)]
+struct SchemaTrieNode {
```
bjchambers: May be better to use https://docs.rs/trie-rs/latest/trie_rs/

PR #467, opened 2025-12-26
feat!: Unify single and multi-item run model

### LINKED ISSUE #465: Simplify execute_flow/execute_batch protocol
Prepare the API / protocol for unifying single and multi-item runs. Specifically, instead of presenting a separate "batch" that has separate runs for each item, change so that a "run" is over one or more items. Doing this first so the implementation can be focused on the executor changes necessary to track tasks.

bjchambers:
This PR flattens the batch/run model by treating multi-item runs as runs with item_count > 1 rather than separate batch entities.

API Changes:
- POST /runs now accepts both single values and arrays via OneOrMany
- Added GET /runs/{run_id}/items endpoint for fetching all item results
- CreateRunResponse includes item_count field

State Store Changes:
- CreateRunParams.input replaced with CreateRunParams.inputs (Vec)
- Added item_count field to RunSummary and RunDetails
- Added database migration 005 for item_count column
- Record item results immediately as they complete in executor
- Use separate run_items table for per-item information

CLI Changes:
- Removed run-batch and submit-batch commands
- Added --inputs option to run and submit for JSONL batch input
- Unified code path in submit.rs for single and batch submissions

Protocol:
- Introduce runs/submit, runs/get methods
- Remove flows/evaluate, flows/get_metadata, flows/submit_batch, flows/get_batch
- Update Rust implementation and Python SDK

Docs:
- Rename flows.md to runs.md with new runs/submit and runs/get docs
- Remove obsolete run-batch.md and submit-batch.md CLI docs
- Update batch-execution.md to use unified run/submit --inputs syntax
- Update protocol index and bidirectional docs for new run methods
- Remove all references to removed flows/evaluate and flows/get_metadata

This is ground work for unifying single-item and multi-item (batch) execution.

This closes #465.

### DIFF
--- a/stepflow-rs/crates/stepflow-cli/src/cli.rs
+++ b/stepflow-rs/crates/stepflow-cli/src/cli.rs
@@ -25,7 +25,6 @@ use crate::{
     infer,
     list_components::OutputFormat,
     repl::run_repl,
-    run::run,
     submit::submit,
     test::TestOptions,
     validate,
@@ -101,14 +100,6 @@ pub enum Command {
         #[command(flatten)]
         input_args: InputArgs,
 
-        /// Path to JSONL file containing multiple inputs (one JSON object per line).
-        ///
-        /// When specified, the workflow is executed once per line in the file.
-        /// Results are written in JSONL format (one result per line).
-        #[arg(long="inputs", value_name = "FILE", value_hint = clap::ValueHint::FilePath,
-              conflicts_with_all = ["input", "input_json", "input_yaml", "stdin_format"])]
-        inputs_path: Option<PathBuf>,
-
         /// Maximum number of concurrent executions (only used with --inputs).
         ///
         /// Defaults to number of inputs if not specified.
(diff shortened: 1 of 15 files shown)


integrations/langflow/tests/helpers/testing/stepflow_binary.py
```diff
@@ -152,6 +153,162 @@ def is_healthy(self) -> bool:
         except Exception:
             return False
 
+    def store_flow(self, flow: dict[str, Any], timeout: float = 30.0) -> str | None:
```
bjchambers: Double check these new test helpers are used / needed. Check whether any of the old helpers can be removed.

stepflow-rs/crates/stepflow-cli/src/cli.rs
```diff
@@ -217,6 +156,20 @@ pub enum Command {
         #[command(flatten)]
         input_args: InputArgs,
 
+        /// Path to JSONL file containing multiple inputs (one JSON object per line).
+        ///
+        /// When specified, the workflow is executed once per line in the file.
+        /// Results are written in JSONL format (one result per line).
+        #[arg(long="inputs", value_name = "FILE", value_hint = clap::ValueHint::FilePath,
```
bjchambers: Should probably be added to the `InputArgs`.

stepflow-rs/crates/stepflow-cli/src/cli.rs
```diff
@@ -480,180 +436,199 @@ impl Cli {
...
-                // Validate workflow and configuration before execution
-                let diagnostics = stepflow_analysis::validate_with_config(
-                    &flow,
-                    &config.plugins,
-                    &config.routing,
-                )
-                .change_context(crate::MainError::ValidationError(
-                    "Validation failed".to_string(),
-                ))?;
+                // Check if we're in batch mode
```
bjchambers: Ideally we can unify this more -- we can make the JSONL loading part of the InputARgs so we can just do ```inputs.load_inputs()``` and then we can determine how to write results based on an `inputs.is_batch()` or some such.

Issue #472, opened 2025-12-29
Refactor: Simplify StepflowExecutor to direct dependencies


bjchambers:
## Summary

Refactor `StepflowExecutor` to remove the circular reference pattern (`Arc::new_cyclic` with `self_weak`) and simplify the dependency structure. The executor should take dependencies directly rather than bundling them in a wrapper struct.

## Background

The current `StepflowExecutor`:
- Uses `Arc::new_cyclic` with `self_weak` for "knot-tying" (flows launching flows)
- Bundles `state_store`, `plugin_router`, and `working_directory`
- Implements the `Context` trait for plugin callbacks
- Acts as both configuration holder and entry point

With the new sub-flow design (where sub-flows are managed inline by `FlowExecutor`), this complexity is no longer needed.

### 1. FlowExecutor takes dependencies directly

```rust
pub struct FlowExecutor {
    // Dependencies passed directly
    state_store: Arc<dyn StateStore>,
    plugin_router: Arc<PluginRouter>,
    working_directory: PathBuf,
    
    // Execution state  
    run_id: Uuid,
    flow: Arc<Flow>,
    flow_id: BlobId,
    state: ItemsState,
    scheduler: Box<dyn Scheduler>,
}
```

### 2. Simple PluginContext for plugin callbacks

```rust
pub struct PluginContext {
    state_store: Arc<dyn StateStore>,
    working_directory: PathBuf,
}

impl Context for PluginContext {
    fn state_store(&self) -> &Arc<dyn StateStore> { ... }
    fn working_directory(&self) -> &Path { ... }
    fn submit_run(&self, params: SubmitRunParams) -> BoxFuture<'_, Result<RunStatus>> {
        // For sub-flows: channel-based communication to executor
        // For now: return error (not supported)
    }
}
```

### 3. Callers create dependencies directly

```rust
// CLI / simple case
let state_store = Arc::new(SqliteStateStore::new(...).await?);
let plugin_router = Arc::new(build_router_from_config(&config)?);

let mut executor = FlowExecutor::new(
    state_store, plugin_router, working_dir,
    run_id, flow, flow_id, inputs, variables,
    Box::new(DepthFirstScheduler::new()),
    max_concurrency,
);

let results = executor.execute_to_completion().await?;

// Server case - stores dependencies in Axum state
struct AppState {
    state_store: Arc<dyn StateStore>,
    plugin_router: Arc<PluginRouter>,
    working_directory: PathBuf,
}
```

[...]

PR #473, opened 2025-12-29
feat: Redesign batch execution with pluggable schedulers

### LINKED ISSUE #347: Split out batch executor
Currently, batch execution happens inside the batches API by creating an executor for each input. We should probably split this out into a separate batch executor that intelligently manages this. We may want to consider work queue, etc.

> Would be good to split out the batch execution logic into a separate executor so we can more intelligently manage things.

_Originally posted by @bjchambers in https://github.com/stepflow-ai/stepflow/pull/345#discussion_r2403704324_

bjchambers:
This PR introduces a new execution architecture that enables cross-item optimization and pluggable scheduling strategies for batch workflows.

## New Architecture

- **FlowExecutor**: Unified executor for single and multi-item runs with async task execution using FuturesUnordered
- **DebugExecutor**: Step-by-step debugging with queue/eval API and support for resuming from persistent state
- **ItemsState**: Cross-item coordinator tracking execution state across all items in a batch with O(1) completion checking
- **ItemState**: Per-item state with BitSet-based dependency tracking

## Pluggable Schedulers

- **Scheduler trait**: Interface for custom execution strategies
- **DepthFirstScheduler**: Prioritizes completing items sequentially (minimizes time-to-first-result)
- **BreadthFirstScheduler**: Groups tasks by step across items (maximizes throughput, enables future batch component invocation)

## Key Improvements

- Schedulers query ItemsState for task status instead of maintaining internal tracking (single source of truth for resume/recovery)
- Atomic counter for O(1) is_all_complete() check in execution loop
- Shared test utilities in testing.rs module
- Comprehensive test coverage including edge cases and failure isolation

## Removed

- Legacy WorkflowExecutor and ExecutionTracker
- Redundant per-item spawning with semaphore-based concurrency

Closes #347

### DIFF
--- a/stepflow-rs/crates/stepflow-execution/src/debug_executor.rs
+++ b/stepflow-rs/crates/stepflow-execution/src/debug_executor.rs
@@ -354,15 +354,6 @@ impl DebugExecutor {
     }
 
     // ========================================================================
-    // Public Query Methods
-    // ========================================================================
-
-    /// Get currently runnable step indices.
-    pub fn get_runnable_step_indices(&self) -> BitSet {
-        self.ready_steps()
-    }
-
-    // ========================================================================
     // Queue/Eval API
     // ========================================================================
 
@@ -790,7 +781,7 @@ mod tests {
         assert!(queued.contains(&"step2".to_string()));
 
         // step1 should be ready (no deps), step2 blocked
-        let ready = debug.get_runnable_step_indices();
+        let ready = debug.ready_steps();
         assert!(ready.contains(0)); // step1
         assert!(!ready.contains(1)); // step2 waiting on step1
     }
@@ -925,7 +916,7 @@ mod tests {
 
         // Queue step2 and run it
         debug.queue_step("step2").await.unwrap();
-        let ready = debug.get_runnable_step_indices();
+        let ready = debug.ready_steps();
         assert!(
             ready.contains(1),
             "step2 should be ready since step1 is already completed"
@@ -1005,7 +996,7 @@ mod tests {
         assert!(!queued.is_empty(), "Debug queue should be recovered");
 
         // step1 should be runnable (recovered from queue)
-        let ready = debug.get_runnable_step_indices();
+        let ready = debug.ready_steps();
         assert!(
             ready.contains(0),
             "step1 should be runnable after queue recovery"
--- a/stepflow-rs/crates/stepflow-execution/src/executor.rs
+++ b/stepflow-rs/crates/stepflow-execution/src/executor.rs
@@ -225,7 +225,7 @@ impl Context for StepflowExecutor {
 
                     log::info!(
                         "Run {run_id} execution completed with {} items",
-                        flow_executor.item_count()
+                        flow_executor.state().item_count()
                     );
                 }
                 .in_span(run_span)
--- a/stepflow-rs/crates/stepflow-execution/src/flow_executor.rs
+++ b/stepflow-rs/crates/stepflow-execution/src/flow_executor.rs
@@ -228,7 +228,7 @@ impl FlowExecutor {
         self.run(None).await?;
 
         // Check for deadlock: run() returned but not complete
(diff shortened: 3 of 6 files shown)


stepflow-rs/crates/stepflow-execution/src/debug_executor.rs
```diff
@@ -0,0 +1,1177 @@
...
+
+        // Create ItemState with workflow input and variables
+        let state = ItemState::new(
+            flow.clone(),
+            input,
+            Arc::new(variables.unwrap_or_default()),
+        );
+
+        // Create workflow-aware execution context
+        let context = ExecutionContext::for_workflow_with_flow(
```
bjchambers: Naming? `for_workflow_with_flow`? Is this still needed? Can it be cleaned up after #472?

stepflow-rs/crates/stepflow-execution/src/debug_executor.rs
```diff
@@ -0,0 +1,1177 @@
...
+        // Queue output dependencies
+        let output_needs = self.flow.output().needed_steps(&self.state);
+        for idx in output_needs.iter() {
+            self.state.add_or_update_needed(idx);
+        }
+
+        // Run until output is resolvable and all needed steps complete
+        loop {
+            // Check if output is ready
+            let output_needs = self.flow.output().needed_steps(&self.state);
```
bjchambers: It feels like a lot of this debug executor is duplicative. I think we may be able to simplify. It seems like the main thing is that while normal execution is "drive everything to compute the output", this is more of a "drive everything to complete what the user has requested". This makes me wonder if we should have the debug executor manage a separate set of "requested" computations, and use that similar to how the `FlowExecutor` drives things. Then all of these are different forms of `drain_requested(...)`. It may not even need to be a queue -- we may be able to just have a method `eval_requested(...)` and then have the different `eval` methods call it.

stepflow-rs/crates/stepflow-execution/src/debug_executor.rs
```diff
@@ -0,0 +1,1177 @@
...
+    // ========================================================================
+
+    /// Recover execution state from the state store and fix any missed status transitions.
+    ///
+    /// This is called internally by [`resume()`](Self::resume) to restore completed steps
+    /// and fix any status mismatches.
+    ///
+    /// # Returns
+    /// The number of step status corrections made during recovery
+    async fn recover_from_state_store(&mut self) -> Result<usize> {
```
bjchambers: This should be two parts -- recreating the `DebugState` from the state store, and then continuing the execution as needed.

stepflow-rs/crates/stepflow-execution/src/executor.rs
```diff
@@ -232,96 +184,78 @@ impl Context for StepflowExecutor {
...
                     } else {
                         stepflow_core::status::ExecutionStatus::Completed
                     };
 
-                    // Update the run status (results already recorded individually)
-                    let _ = executor_clone
-                        .state_store()
+                    // Update the run status
+                    let _ = state_store_clone
                         .update_run_status(run_id, final_status)
```
bjchambers: Updating the run status on completion should be part of the FlowExecutor.

stepflow-rs/crates/stepflow-execution/src/flow_executor.rs
```diff
@@ -0,0 +1,871 @@
...
+        }
+    }
+
+    /// Execute all items to completion.
+    ///
+    /// Returns a vector of FlowResults, one per item, in item index order.
+    /// Results are recorded to the state store as items complete.
+    ///
+    /// Tasks are executed concurrently up to `max_in_flight` at a time.
+    pub async fn execute_to_completion(&mut self) -> Result<Vec<FlowResult>> {
```
bjchambers: We could allow this to be a stream of results and include the`ItemIndex` in the result. Then, we wouldn't need the order to match, nor would we need to collect all the results.

PR #481, opened 2026-01-06
refactor: Rename stepflow-py package to stepflow-server

kerinin:
Restructure Python SDK to use a UV workspace layout with stepflow-server as the main component server package. This prepares the way for additional packages. The `stepflow-py` package name will eventually be used for a catchall including this and several others.

Changes:
- Move sdks/python/src/stepflow_py/ to sdks/python/stepflow-server/src/stepflow_server/
- Move tests to sdks/python/stepflow-server/tests/
- Update all imports from stepflow_py to stepflow_server
- Update pyproject.toml for UV workspace structure
- Update stepflow-config.yml files to reference new package path
- Update example Python files with new imports
- Update CLAUDE.md documentation
- Update langflow integration imports

### DIFF
--- a/sdks/python/README.md
+++ b/sdks/python/README.md
@@ -6,7 +6,7 @@ Python SDK for building Stepflow components and workflows.
 
 ```bash
 # Install from source
-uv add stepflow-py
+uv add stepflow-worker
 ```
 
 ## Configuration
@@ -15,7 +15,7 @@ The SDK is configured via environment variables:
 
 | Variable | Description | Default |
 |----------|-------------|---------|
-| `STEPFLOW_SERVICE_NAME` | Service name for observability | `stepflow-python` |
+| `STEPFLOW_SERVICE_NAME` | Service name for observability | `stepflow-workerthon` |
 | `STEPFLOW_LOG_LEVEL` | Log level (DEBUG, INFO, WARNING, ERROR) | `INFO` |
 | `STEPFLOW_LOG_DESTINATION` | Log destination (stderr, file, otlp, comma-separated) | `otlp` if OTLP endpoint set, else `stderr` |
 | `STEPFLOW_LOG_FILE` | File path for file logging | - |
@@ -26,7 +26,7 @@ The SDK is configured via environment variables:
 ```bash
 STEPFLOW_SERVICE_NAME=my-python-components \
 STEPFLOW_OTLP_ENDPOINT=http://localhost:4317 \
-uv run stepflow_py
+uv run stepflow_worker
 ```
 
 **Example with OTLP + stderr logging:**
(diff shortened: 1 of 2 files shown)


sdks/python/stepflow-server/README.md
```diff
@@ -0,0 +1,44 @@
+# stepflow-server
```
bjchambers: Thoughts on naming? I think this could make sense, but I'm wondering if `server` is confusing.

It makes sense relative MCP where you have the "client" (the agent) using the MCP "server" (the tools). But, I could see it getting muddied with the Stepflow orchestrator (server). We may need to cleanup the naming over there.

Other options I've thought of but don't think are necessarily better: `stepflow-sdk` (the SDK for authoring component servers), `stepflow-components` (provides components), `stepflow-component-server` (longer, but makes it clear what it is).

If any of those seem to be clearer, we can discuss.

examples/production-model-serving/text_models_server.py
```diff
@@ -34,11 +34,11 @@
...
 
 try:
-    from stepflow_py import StepflowStdioServer, StepflowContext
-    from stepflow_py.server import StepflowServer
+    from stepflow_server import StepflowStdioServer, StepflowContext
+    from stepflow_server.server import StepflowServer
 
     try:
-        from stepflow_py.http_server import StepflowHttpServer
+        from stepflow_server.http_server import StepflowHttpServer
```
bjchambers: Search for places with strings containing `stepflow-py`. Some packages (like this) capture the import error and say "stepflow SDK not available. install stepflow-py to run". we should update thatso these examples are friendly.

.github/workflows/release_python.yml
```diff
@@ -45,11 +45,11 @@ jobs:
             echo "Tag: ${{ github.event.client_payload.tag }}"
           else
             echo "version=${{ inputs.version }}" >> $GITHUB_OUTPUT
-            echo "tag=stepflow-py-${{ inputs.version }}" >> $GITHUB_OUTPUT
+            echo "tag=stepflow-worker-${{ inputs.version }}" >> $GITHUB_OUTPUT
             echo "skip_tag=${{ inputs.skip_tag_creation }}" >> $GITHUB_OUTPUT
             echo "Source: manual workflow dispatch"
             echo "Version: ${{ inputs.version }}"
-            echo "Tag: stepflow-py-${{ inputs.version }}"
+            echo "Tag: stepflow-worker-${{ inputs.version }}"
```
bjchambers: Do we want to support release the different python packages separately? That could get hairy. What if we just have the tag be stepflow-py and release all python packages at that version?

We can maybe revert these changes and work on the release process separately, if that is easier.

examples/custom-python-components/custom_server.py
```diff
@@ -27,7 +27,7 @@
 import asyncio
 import sys
 
-from stepflow_py import StepflowServer, StepflowContext
+from stepflow_worker import StepflowServer, StepflowContext
```
bjchambers: This looks good. I wonder if we should plan to rename `StepflowServer` to `StepflowWorker` in the future (before we do a release of this change). Doesn't need to be in this PR since this is just trying to make space for other things.

I can see a case that this is the "Worker Server" and we could emphasize either or both of those. StepflowServer may be fine.

kerinin: Good catch - agree this should be changed too. Will do in another PR

.github/workflows/release_python.yml
bjchambers: Makes sense -- let's revert the changes to the release scripts in this PR then and do a separate PR to focus on it (better to have a known working state and then change it rather than have partial changes).

PR #482, opened 2026-01-06
feat: Add Flow::latest_mut and fix WorkflowOverrides serialization

kerinin:
Rust API changes needed for OpenAPI-compatible Python client:

- Add latest_mut() method to Flow enum for mutable access to FlowV1
- Remove #[serde(transparent)] from WorkflowOverrides to fix serialization
- Update JSON schemas to reflect Rust type changes
- Add OpenAPI schema for HTTP API documentation


schemas/flow.json
```diff
@@ -269,146 +193,8 @@
       }
     },
     "ValueExpr": {
-      "oneOf": [
```
bjchambers: Why did this change? I'm pretty sure VlaueExpr is still a oneof (it's an enum) and it seems like ti has more details when we break it out like that.

bjchambers:
And diagnostic message we could probably edit the schema to use a named type for each diagnostic and/or use a description.

PR #483, opened 2026-01-07
feat: 428 k8s refactor

zznate:
Move k8s plumbing for langflow component+load balancer demo into repo. This is mostly copied over from stepflow-langflow repository with a few small changes.


zznate:
Merging this cause it's isolated and we have a lot of PRs in flight.

bjchambers:
It feels like we shouldn't have this in the top level k8s directory. I feel like that pattern is for projects that are primarily built as a service -- "this is the k8s for the project". I think Stepflow is a *piece* you deploy to your k8s. We can provide an *example* showing how to do that, but we don't want to present it as "to use Stepflow, you must use our k8s and it has this tracing setup, etc.". It feels like this should be moved to `examples/k8s` or `examples/production` with `examples/production/k8s` having the k8s directory structure and `examples/production/docker` containing docker files of unreleased things, etc.

PR #486, opened 2026-01-07
refactor!: Replace STDIO transport with unified HTTP transport

bjchambers:
BREAKING CHANGE: Remove STDIO-based component communication protocol in favor of HTTP-only transport with subprocess and remote URL support.

## Transport Changes

Remove STDIO transport:
- Delete stdio/client.rs, stdio/launcher.rs, and stdio/recv_message_loop.rs
- Remove STDIO-specific protocol handling and message loop
- Eliminate bidirectional STDIO complexity and framing overhead

Unify on HTTP transport:
- Single transport for both managed subprocesses and remote URLs
- Subprocesses now use HTTP via subprocess/launcher.rs
- Launch subprocesses that print port to stdout, then communicate via HTTP
- Supports both subprocess mode (command + args) and remote mode (url)

## Python SDK Changes

Update to HTTP-only:
- Replace stdio_server.py with http_server.py as primary interface
- Update StepflowHttpServer to use FastAPI/Uvicorn
- Print port announcement {"port": N} to stdout for subprocess discovery
- Remove STDIO server implementation and tests

Update examples and configs:
- Change all stepflow-config.yml to use HTTP transport
- Update Python component servers to use StepflowHttpServer
- Update Langflow integration for HTTP-only mode

## Subprocess Management Improvements

Fix subprocess cleanup with three-part solution:

1. Break circular reference with Weak<dyn Context>:
   - HttpClient stores Weak<dyn Context> instead of Arc
   - Breaks cycle: executor → plugin → client → context → executor
   - Allows Drop to run when server shuts down

2. Add SIGTERM/SIGINT signal handling:
   - Graceful shutdown on signal reception
   - Allows Drop implementations to run instead of abrupt termination
   - Server logs shutdown progress for debugging

3. Use Unix process groups for grandchildren:
   - Spawn subprocesses in new process group via setpgid
   - Kill entire process group in Drop (SIGTERM)
   - Handles cases like uv spawning Python servers
   - Prevents orphaned grandchild processes

Without all three pieces, subprocesses would leak:
- Without Weak: Circular reference prevents Drop from ever running
- Without signals: Process killed immediately, no cleanup code runs
- Without process groups: kill_on_drop only kills direct child, not grandchildren

Add nix dependency for Unix signal handling (killpg, setpgid).

## Protocol Updates

Simplify with single transport:
- Remove transport-specific error types
- Consolidate connection handling in StepflowPlugin
- Single code path for HTTP client creation
- Cleaner plugin initialization logic

Remove deprecated/unused code:
- Remove shutdown() method from Plugin trait (now handled by Drop)
- Remove explicit plugin cleanup (automatic via Drop chain)
- Simplify executor lifecycle management

## Documentation Updates

Update all documentation for HTTP-only:
- Component server implementation guides
- Configuration examples
- Deployment documentation
- Best practices for testing

Fixes orphaned subprocess issue that caused test flakiness in Langflow integration.

### DIFF
--- a/integrations/langflow/src/stepflow_langflow_integration/cli/main.py
+++ b/integrations/langflow/src/stepflow_langflow_integration/cli/main.py
@@ -185,9 +185,7 @@ def analyze(input_file: Path):
 
 
 @main.command()
-@click.option(
-    "--host", default="localhost", help="Server host"
-)
+@click.option("--host", default="localhost", help="Server host")
 @click.option("--port", default=0, help="Server port (0 for auto-assign)")
 @click.option("--protocol-prefix", default="langflow", help="Protocol prefix")
 def serve(


examples/custom-python-components/custom_server.py
```diff
@@ -24,13 +24,16 @@
...
+import sys
+
+from stepflow_py import StepflowHttpServer, StepflowServer, StepflowContext
 import msgspec
 from typing import List, Optional
-import asyncio
 
 # Create the server
-server = StepflowStdioServer()
+_server = StepflowServer()
```
bjchambers: Since there is only the stepflow server, we should rename these classes and combine them.

examples/langchain/langchain_server.py
```diff
@@ -151,22 +155,33 @@ async def invoke_named(input: InvokeNamedInput, context: StepflowContext) -> dic
 
 
 if __name__ == "__main__":
-    print("LangChain Stepflow Integration Examples")
-    print("======================================")
-    print()
+    import sys
+
+    print("LangChain Stepflow Integration Examples", file=sys.stderr)
```
bjchambers: Can print to stdout now, no need to change to stderr.

PR #491, opened 2026-01-08
feat: Add stepflow-api Python client generated from OpenAPI spec

kerinin:
Add a Python API client package generated using openapi-generator:
- New stepflow-api package with Pydantic models and typed API methods
- Generator script (scripts/generate_api_client.py) with docker/podman support
- Stores openapi.json schema for reproducible regeneration

The generator uses openapi-generator (via container) which properly handles recursive types, generating List[ValueExpr] and Dict[str, ValueExpr] for nested expression structures.

### DIFF
--- a/.licensure.yml
+++ b/.licensure.yml
@@ -1,6 +1,7 @@
 excludes:
   - docs/.*
   - stepflow-ui/.*
+  - sdks/python/stepflow-api/.*
   - .*\.ya?ml
   - .*\.snap
   - .*\.toml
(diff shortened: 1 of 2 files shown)


sdks/python/scripts/generate_api_client.py
```diff
@@ -0,0 +1,446 @@
+#!/usr/bin/env python3
```
kerinin: This file is 99% of the change here - the rest is just setting up the package and generated code. A follow-up PR adds package publishing and other integrations.

PR #492, opened 2026-01-08
feat: Enable running Stepflow workflows from Python scripts

kerinin:
Add packages for executing workflows directly from Python code:

- stepflow-runtime: Run workflows locally with bundled Stepflow binary
  - Zero-configuration embedded runtime
  - Automatic binary management and health checks

- stepflow-client: Execute workflows on remote Stepflow servers
  - Built on generated OpenAPI client for type safety
  - Async HTTP client with full API coverage

- stepflow-core: Shared types implementing StepflowExecutor protocol
  - FlowResult, RestartPolicy, and other core types
  - Common interface for local and remote execution

- stepflow-py: Unified package re-exporting all components
  - Single import for both client and server functionality

Includes CI workflow for cross-platform builds and PyPI publishing.

### DIFF
--- a/sdks/python/stepflow-py/src/stepflow_py/__init__.py
+++ b/sdks/python/stepflow-py/src/stepflow_py/__init__.py
@@ -90,7 +90,6 @@ from stepflow_worker import (
     FlowBuilder,
     StepflowContext,
     StepflowServer,
-    StepflowStdioServer,
     Value,
 )
 
@@ -120,7 +119,6 @@ __all__ = [
     "StepflowClientError",
     # Component Server
     "StepflowServer",
-    "StepflowStdioServer",
     "StepflowContext",
     # Workflow Building
     "FlowBuilder",


sdks/python/stepflow-client/tests/test_client.py
```diff
@@ -0,0 +1,485 @@
...
+        assert str(error) == "Bad request"
+        assert error.status_code == 400
+        assert error.details["errors"] == ["field1 required", "field2 invalid"]
+
+
+class TestStepflowClientFlowLoading:
+    """Test flow loading from different sources."""
+
+    async def test_store_flow_from_dict(self):
+        """Test storing a flow from a dictionary."""
```
kerinin: Does this actually test anything meaningful?

sdks/python/stepflow-client/tests/test_client.py
```diff
@@ -0,0 +1,485 @@
...
+            new_callable=AsyncMock,
+            return_value=mock_response,
+        ) as mock_create:
+            await client.create_run(flow_id="flow123", input=[{"x": 1}, {"x": 2}])
+
+            call_args = mock_create.call_args
+            request = call_args.kwargs["body"]
+            assert request.input_ == [{"x": 1}, {"x": 2}]
+
+    async def test_error_response_raises_exception(self):
```
kerinin: Same for all these tests with mocking - what is actually being tested here? At minimum please update comments to clarify what is being tested, given the use of mocks.

kerinin:
We could probably consolidate some of the packages, but the reasoning for them was basically:
* Stepflow-worker: Separate becuase worker implementors are likely to not need the client functionality
* stepflow-runtime: Separate because it packages the compiled runtime which is ~40Mb, and isn't needed if you're just talking to a remote API.

The other packages sort of just fell out of the pattern of separating things. If we wanted to minimize the package count we could go with:

stepflow-worker - only import this if you're building workers
stepflow-py - current api + core + client - import this if you're calling stepflow
stepflow-runtime - import this if you're executing in-process

Let's chat : )

Issue #522, opened 2026-01-16
Add a docling component for deeper integration both from treating docling as a top level component and from within Langflow flows


zznate:
Add Docling as a first-class Stepflow component namespace, enabling document processing capabilities independent of Langflow while maintaining seamless integration when Langflow flows contain Docling components.

## Proposed Components

- `/docling/convert` — Document processing (PDF, DOCX, images → structured data)
- `/docling/chunk` — Semantic chunking for RAG pipelines
- `/docling/export` — Format conversion (Markdown, HTML, JSON)

## Key Benefits

- Use Docling directly in any Stepflow flow, not just Langflow
- Unified observability through Stepflow's OTel stack
- Independent scaling of document processing workloads
- GPU routing for heavy OCR/table extraction

## Details

See full proposal: https://github.com/stepflow-ai/stepflow/blob/522-docling-integration/docs/proposals/docling-integration.md

PR #541, opened 2026-01-26
feat: Introduces Docling integration component into integrations patt…

zznate:
…ern similar to langflow. Partially addresses #522. Additional integration tests to perform

### DIFF
--- a/integrations/langflow/src/stepflow_langflow_integration/converter/known_components.py
+++ b/integrations/langflow/src/stepflow_langflow_integration/converter/known_components.py
@@ -66,12 +66,18 @@ KNOWN_COMPONENTS: dict[str, KnownComponent] = {
     # from custom code.
     #
     # Docling components (lfx)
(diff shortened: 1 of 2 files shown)


integrations/langflow/src/stepflow_langflow_integration/converter/known_components.py
```diff
@@ -64,13 +64,26 @@ class KnownComponent:
     # Note: PromptComponent (langflow.components.prompts.prompt) is NOT included
     # because the module doesn't exist in the lfx package - it must be compiled
     # from custom code.
+    #
+    # Docling components (lfx)
+    # Note: DoclingRemote uses custom_components.docling_serve which requires
```
bjchambers: Can you elaborate on *why* `DoclingRemote` uses `custom_components.docling_serve` It seems like it should be possible for it to be hashed?

PR #571, opened 2026-02-02
refactor: Consolidate ExecutionContext into RunContext

### LINKED ISSUE #505: Simplify RunContext vs. ExecutionContext?
From PR #502 

> Open an issue to look at the relationship between RunContext and ExecutionContext. I believe one should be removable. Also, should we just have a RunContext::new(...) instead of the for_root(...).with_submitter(...) pattern?

bjchambers:
Simplify the plugin execution API by merging ExecutionContext into RunContext. The unified RunContext now carries all execution state: run hierarchy (run_id, root_run_id), flow and flow_id, environment, and subflow submission.

Key changes:
- Remove ExecutionContext, use RunContext throughout
- Plugin::execute takes run_context and step separately
- StepRunner constructor simplified to just step_index, input, run_context
- Update all plugins and handlers to use the new API

Closes #505

### DIFF
--- a/stepflow-rs/crates/stepflow-builtins/src/plugin.rs
+++ b/stepflow-rs/crates/stepflow-builtins/src/plugin.rs
@@ -15,13 +15,14 @@ use std::sync::Arc;
 use crate::{BuiltinComponent as _, registry};
 use error_stack::ResultExt as _;
 use serde::{Deserialize, Serialize};
+use stepflow_core::workflow::StepId;
 use stepflow_core::{
     FlowResult,
     component::ComponentInfo,
     workflow::{Component, ValueRef},
 };
 use stepflow_plugin::{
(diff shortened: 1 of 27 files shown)


stepflow-rs/crates/stepflow-builtins/src/plugin.rs
```diff
@@ -75,12 +76,13 @@ impl Plugin for Builtins {
     async fn execute(
         &self,
         component: &Component,
-        context: ExecutionContext,
+        run_context: &Arc<RunContext>,
+        step: Option<&StepId>,
```
zznate: Why is step optional? I'm not seeing why we would run a plugin without a step? More of a general question than a cristicism - i'm probably not seeing something. Though I *can* see how this is useful for testing.

zznate (approved):
No changes - just a couple of clarifications and a copilot comment suggestion.

PR #577, opened 2026-02-04
chore: Updates and tweaks for getting latest openrag ingestion flow working on k8s example architecture

zznate:
Not quite ready for review yet.

### DIFF
--- a/integrations/langflow/src/stepflow_langflow_integration/executor/langflow_server.py
+++ b/integrations/langflow/src/stepflow_langflow_integration/executor/langflow_server.py
@@ -93,12 +93,28 @@ class StepflowLangflowServer:
 
     def run(self, *args, **kwargs) -> None:
         """Run the component server."""
+        import os
+
         # Apply nest_asyncio to allow nested event loops
         # This is needed because Langflow components may call asyncio.run()
         # from within an already-running event loop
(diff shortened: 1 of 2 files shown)


docker/langflow-component-server/components.py
```diff
@@ -42,6 +41,11 @@
 
 def main():
     """Main entry point for HTTP server mode."""
+    # Initialize observability (tracing, logging) before anything else
+    from stepflow_py.worker.observability import setup_observability
```
bjchambers: It feels like we should move this into the stepflow_langflow_integration.executor.langflow_server -- eg., we should have this in the python project, rather than adding code to the docker image script.

sdks/python/stepflow-py/src/stepflow_py/worker/server.py
```diff
@@ -497,9 +497,18 @@ async def _handle_component_execute(
                 },
             ):
                 # Parse input using component's input type
-                input_value: Any = msgspec.convert(
-                    params.input, type=component.input_type
-                )
+                # Handle types that msgspec doesn't support (dict, Any, object)
```
bjchambers: This seems a little suspect. File an issue to investigate / improve?

PR #585, opened 2026-02-05
feat: Split StateStore and add recovery architecture

### LINKED ISSUE #462: Add conformance tests for state store specification
We should have a suite of tests that we can run on any state store implementation to verify basic characteristics.

- [ ] Run / hierarchy of runs
- [ ] Items in runs

### LINKED ISSUE #579: Split up StateStore and introduce leases/recovery
Create the new traits for #578:

1. Split StateStore into `MetadataStore`, `ExecutionJournal` and `LeaseManager`.
2. Introduce recovery interface using `LeaseManager`.

Note: `BlobStore` to be introduced in a follow-up.

bjchambers:
This PR adds comprehensive persistence and recovery architecture including:
    
**Storage Configuration Refactoring:**
- Renamed `StateStoreConfig` → `StorageConfig` with field `storageConfig` in YAML
- Split single `StateStore` trait into three focused traits: `MetadataStore`, `BlobStore`, `ExecutionJournal`
- Added expanded storage config format allowing separate backends per store type:
  ```yaml
  # Simple form (all stores share one backend)
  storageConfig:
    type: sqlite
    databaseUrl: "sqlite:workflow_state.db"
    
  # Expanded form (individual configs per store)
  storageConfig:
    metadata:
      type: sqlite
      databaseUrl: "sqlite:workflow_state.db"
    blobs:
      type: sqlite
      databaseUrl: "sqlite:workflow_state.db"
    journal:
      type: inMemory
  ```
- Smart deduplication: identical configs share a single backend instance
    
**Journal Compliance Tests (ExecutionJournal trait):**
- Monotonic sequence numbers, read/write operations
- Checkpoint and compaction behavior
- Subflow shared journal (keyed by root_run_id)
- Event serialization round-trips
    
**Metadata Compliance Tests (MetadataStore trait):**
- Blob storage with deduplication
- Run lifecycle (create, status transitions, idempotency)
- Item results recording and retrieval
- Run hierarchy (root_run_id, parent_run_id filters)
- Recovery via list_pending_runs
    
**Recovery Architecture Simplifications:**
- Removed `executing` BitSet from ItemState (executor tracks in-flight tasks)
- Renamed `ready_steps()` to `schedulable_steps()` for clarity
- Removed `apply_events` wrapper (callers use for_each directly)
- Event replay determinism tested directly in unit tests
    
**Also adds:**
- `roots_only` filter for listing only top-level runs
- Fixed `list_pending_runs` query in SQLite to join with run_items
- Fixed in-memory sequence number preservation after compaction
- Journal buffering with explxcit flush
    
Closes #462
Closes #579

### DIFF
--- a/stepflow-rs/crates/stepflow-execution/src/recovery.rs
+++ b/stepflow-rs/crates/stepflow-execution/src/recovery.rs
@@ -323,12 +323,14 @@ mod tests {
     /// Helper to create a simple test flow.
     fn create_test_flow() -> stepflow_core::workflow::Flow {
         FlowBuilder::test_flow()
-            .steps(vec![StepBuilder::new("step0")
-                .component("/mock/test")
-                .input(ValueExpr::Input {
-                    input: Default::default(),
-                })
-                .build()])
+            .steps(vec![
+                StepBuilder::new("step0")
+                    .component("/mock/test")
+                    .input(ValueExpr::Input {
+                        input: Default::default(),
+                    })
+                    .build(),
+            ])
             .output(ValueExpr::Step {
                 step: "step0".to_string(),
                 path: Default::default(),
--- a/stepflow-rs/crates/stepflow-server/src/api/runs.rs
+++ b/stepflow-rs/crates/stepflow-server/src/api/runs.rs
@@ -443,7 +443,11 @@ pub async fn get_run_steps(
                 error_stack::report!(ServerError::InvalidRequest(format!(
                     "Item index {} out of bounds (run has {} items)",
                     item_index,
-                    execution.item_details.as_ref().map(|i| i.len()).unwrap_or(0)
+                    execution
+                        .item_details
+                        .as_ref()
+                        .map(|i| i.len())
+                        .unwrap_or(0)
                 )))
             })?
     } else {
@@ -492,7 +496,9 @@ fn aggregate_step_statuses(
 
     for item in items {
         for step in &item.steps {
-            let current = aggregated.entry(step.step_id.clone()).or_insert(step.status);
+            let current = aggregated
+                .entry(step.step_id.clone())
+                .or_insert(step.status);
             // Update if new status has higher precedence (is "worse")
             if status_precedence(step.status) > status_precedence(*current) {
                 *current = step.status;


schemas/stepflow-config.json
```diff
@@ -103,6 +111,49 @@
         }
       ]
     },
+    "LeaseManagerConfig": {
+      "oneOf": [
```
bjchambers: This schema seems incorrect (and the code it generates also seems incorrect). I think it needs a description with a `# Title` so the python code generates a name (eg., `NoLeaseManager` or `LeaseManagerNone` or something like that). Probably `NoLeaseManager` or `NoopLeaseManager` so we could have `EtcdLeaseManager`, etc.

sdks/python/stepflow-py/src/stepflow_py/config/_generated_config.py
```diff
@@ -286,3 +298,12 @@ class StepflowConfig(RoutingConfig, kw_only=True):
         ]
         | None
     ) = None
+    leaseManager: (
```
bjchambers: Need to regenerate to get recovery config generated as well.

stepflow-rs/crates/stepflow-execution/src/state/item_state.rs
```diff
@@ -236,16 +232,19 @@ impl ItemState {
         newly_unblocked
     }
 
-    /// Get the set of steps that are ready to execute.
+    /// Get the set of steps that are schedulable for execution.
     ///
-    /// A step is ready if it is needed, not executing, not completed,
+    /// A step is schedulable if it is needed, not completed,
     /// and has no waiting dependencies.
```
bjchambers: A "waiting dependency" seems odd. It seems like it would be better to say:

1. All necessary dependencies are completed, or
2. All steps it is waiting on are complete, or
3. It is not waiting on any incomplete steps (or, not waiting on any steps).

stepflow-rs/crates/stepflow-execution/src/executor.rs
```diff
@@ -80,14 +82,55 @@ pub async fn submit_run(
         .await
         .change_context(ExecutionError::StateStoreError)?;
 
+    // Journal: Record run creation
+    if let Some(journal) = env.execution_journal() {
```
bjchambers: Would be easier if we *always* had a journal, even if it was a noop. Let's make this required in the environment, and eliminate these conditionals.

stepflow-rs/crates/stepflow-execution/src/executor.rs
```diff
@@ -80,14 +82,55 @@ pub async fn submit_run(
...
+            run_id, // For root runs, root_run_id == run_id
+            JournalEvent::RunCreated {
+                flow_id: flow_id.clone(),
+                inputs: inputs.clone(),
+                variables: params.variables.clone().unwrap_or_default(),
+                parent_run_id: None,
+            },
+        );
+        if let Err(e) = journal.append(entry).await {
+            log::warn!(
```
bjchambers: This should probably be a *fatal* error and cause us to abandon this attempt on the run. Rationale -- if we don't journal results, then anything that has side effects may be retried with different values if steps are non deterministic. Without hte journal we won't be able to recover.

This is true here and in other places we journal -- it is not best effort, it is essential.

That said, we may want a way to make this non-blocking -- if the append just returns and there is a separate `journal.flush().await` or `journal.sync().await`, then we can make sure to synchronize at necessary points. For instance, if we have steps that complete all ta the same time, we may want to append them all before we flush the journal.

The only requirement should be that we flush before starting anything new which may have side-effects.

stepflow-rs/crates/stepflow-execution/src/flow_executor.rs
```diff
@@ -109,6 +109,23 @@ impl FlowExecutor {
         self.runs.get(&self.root_run_id).expect("root run exists")
     }
 
+    /// Write a journal entry if journalling is enabled.
+    ///
+    /// This is a fire-and-forget operation - journal write failures are logged
+    /// but don't halt execution. The journal is for recovery, not for correctness.
```
bjchambers: See previous comments. Appending to the journal should return quickly (and be critical). Flushing the journal should be awaited before launching steps.

stepflow-rs/crates/stepflow-execution/src/recovery.rs
```diff
@@ -0,0 +1,298 @@
...
+///
+/// # Arguments
+/// * `env` - The Stepflow environment
+/// * `lease_manager` - The lease manager for claiming runs
+/// * `orchestrator_id` - This orchestrator's ID
+/// * `limit` - Maximum number of runs to recover
+///
+/// # Returns
+/// A `RecoveryResult` describing what was recovered
+pub async fn recover_pending_runs(
```
bjchambers: This is misleadingly named. It doesn't really recover *pending* runs. It recovers *orphaned* runs (those that aren't currently claimed by a running orchestrator).

stepflow-rs/crates/stepflow-state-sql/src/lib.rs
```diff
@@ -132,4 +136,488 @@ mod tests {
         // Should get the same ID (deduplication)
         assert_eq!(blob_id1, blob_id2);
     }
+
+    // =========================================================================
+    // ExecutionJournal Tests
```
bjchambers: Are these tests duplicative with the execution journal compliance tests? If not, can we move them to a:

```
#[cfg(test)]
mod sql_execution_journal_tests;
```

and have a separate `sql_execution_journal_tests.rs`?

stepflow-rs/crates/stepflow-state-sql/src/migrations.rs
```diff
@@ -25,6 +25,12 @@ pub async fn run_migrations(pool: &SqlitePool) -> Result<(), StateError> {
     })
     .await?;
 
+    // Apply journal tables migration
+    apply_migration(pool, "002_create_journal_tables", || {
+        create_journal_tables(pool)
```
bjchambers: Should just modify the 001 migration (no backwards compatibility requirements yet). Should also delete any unecessary tables, and consider splitting the migrations up to configure only the tables necessary for the functinoality the SQL store will be used for (eg., only create the journal tables if we're using it for journalling, etc.)

stepflow-rs/crates/stepflow-state/src/execution_journal.rs
```diff
@@ -0,0 +1,319 @@
...
+    ) -> BoxFuture<'_, error_stack::Result<Option<SequenceNumber>, StateError>>;
+
+    /// Remove journal entries before the checkpoint.
+    ///
+    /// This is used for garbage collection after entries have been
+    /// durably synced to MetadataStore.
+    ///
+    /// # Arguments
+    /// * `root_run_id` - The root run to compact
+    fn compact(&self, root_run_id: Uuid) -> BoxFuture<'_, error_stack::Result<(), StateError>>;
```
bjchambers: This seems like it should take the checkpoint that we can compact up to. It also seems like it isn't consistent with the whole journal usage. I think the journal should roughly be:

1. Append to it when an event happens. This may immediately write or it may just queue the write.
2. Flush the journal. Flushing returns a sequence number indicating the point it flushed to (this may be a checkpoint, offset, etc.). This allows batching journal writes, but create "barriers" when we need to flush to know that something is durable committed.

For now, we could probably omit the compaction/checkpointing logic, since it isn't fully fleshed out. Let's drop that functionality for the time being.

Instead, we should document that:

1. Sequence numbers don't need to start at `0` (consider a shared journal). In the future, we may need to allow the sequence number or offset to be implementation defined.
2. In the metadata store, we should store the sequence number of offset of the *first* event in the journal for a given run.

scripts/check-rust.sh
```diff
@@ -40,27 +40,27 @@ cd "$PROJECT_ROOT/stepflow-rs"
 # RUST STYLE & QUALITY CHECKS
 # =============================================================================
 
-run_check "Formatting" --fix "cargo fmt" cargo fmt --check
+run_check "Formatting" --fix "cargo fmt" cargo fmt --check || true
```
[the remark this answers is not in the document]
bjchambers: This is intentional - the `|| true` allows all checks to run even if one fails, so you see ALL failures at once rather than just the first. Failures are tracked in the `FAILED_CHECKS` array and `print_summary` returns a non-zero exit code if any check failed. See the NOTE comment at line 38.

PR #629, opened 2026-02-17
feat: Replace the StoredBlob JSON wrapper with a trailing-metadata

### LINKED ISSUE #624: Implement direct byte access for local files
Currently, the local file store blob storage uses a file named after the blob ID which contains JSON holding the file name and the base64 encoded file content. We should change this to some mechanism that allows direct byte access:

1. We could store the file directly and have an adjacent file containing the metadata
2. We could store a header followed by the raw bytes of the file for instance:

```
{
  file_name: <name>
  content_offset: <usize indicating where the file starts>,
}
FILECONTENTHERE
...
```

From https://github.com/stepflow-ai/stepflow/pull/621#discussion_r2805488375

zznate:
Closes #624

File format in this commit:

```
  [CONTENT BYTES][METADATA JSON][CONTENT_LENGTH (4 bytes, LE u32)]
```

Key implementation details:
  1. Removed StoredBlob, replaced with BlobFileMetadata — a compact struct serialized as trailing JSON metadata containing blob_type and optional filename.
  2. Extensionless paths — blob_path() now returns <dir>/ab/<hash> (no .json). Temp files use .tmp suffix.
  3. put_blob rewritten:
    - Binary blobs: base64-decoded to raw bytes
    - Data/Flow blobs: compact JSON serialization (not pretty-printed)
    - Atomic write via .tmp → rename
  4. get_blob_opt rewritten:
    - Validates file >= 4 bytes, reads content_length from footer
    - Validates content_length doesn't exceed available space
    - Splits content/metadata, deserializes metadata JSON
    - Binary blobs reconstructed directly from raw bytes (no base64 round-trip)

### DIFF
--- a/stepflow-rs/crates/stepflow-state/src/filesystem_blob_store.rs
+++ b/stepflow-rs/crates/stepflow-state/src/filesystem_blob_store.rs
@@ -20,13 +20,13 @@
 //! Each blob file uses a trailing-metadata layout:
 //!
 //! ```text
-//! [CONTENT BYTES (variable)]  [METADATA JSON (variable)]  [CONTENT_LENGTH (4 bytes, LE u32)]
+//! [CONTENT BYTES (variable)]  [METADATA JSON (variable)]  [CONTENT_LENGTH (8 bytes, LE u64)]
 //! ```
 //!
 //! Reading strategy:
-//! 1. Read last 4 bytes as `u32` (little-endian) → `content_length`
+//! 1. Read last 8 bytes as `u64` (little-endian) → `content_length`
 //! 2. Content is bytes `0..content_length`
-//! 3. Metadata JSON is bytes `content_length..(file_size - 4)`
+//! 3. Metadata JSON is bytes `content_length..(file_size - 8)`
 //!
 //! ## Storage Modes
 //!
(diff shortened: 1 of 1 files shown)


stepflow-rs/crates/stepflow-state/src/filesystem_blob_store.rs
```diff
@@ -129,18 +144,48 @@ impl crate::BlobStore for FilesystemBlobStore {
...
+            buf.extend_from_slice(&content_bytes);
+            buf.extend_from_slice(&metadata_bytes);
+            buf.extend_from_slice(&content_length.to_le_bytes());
+
+            // Write atomically: write to .tmp then rename
+            let temp_path = path.with_extension("tmp");
+            std::fs::write(&temp_path, &buf)
                 .change_context(StateError::Internal)
                 .attach_printable_lazy(|| {
                     format!("Failed to write blob file: {}", temp_path.display())
```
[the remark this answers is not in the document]
bjchambers: This may need to have a nonce/uuid included to avoid this problem. Or, tolerating the rename failure as described here.

stepflow-rs/crates/stepflow-state/src/filesystem_blob_store.rs
bjchambers: We may want to change the API to allow a separate path for putting binary data.

Currently, the API server accepts either raw binary data or JSON data. This means that in order to *call* this, it has to do a base64 encoding which is immediately undone. We should change the API so that it can call method to put raw data.

It *may* mean we're better off having this take the bytes or byte stream to write rather than a `data: ValueRef`. Then, instead of having a "binary" API and a "JSON" API, we'd just move the JSON->binary up. This allows us to support "binary->binary", and "JSON->binary" at the API level, and the blob store trait only has to support binary.

bjchambers: Actually we could maybe keep the two methods if we thought some backends may prefer storing base64. Then the api could call the one matching what it receives and the backend could decode/encode if preferred.

But, if we have to choose one format to support it seems like binary is most likely to avoid extra complexity in clients, etc.

PR #635, opened 2026-02-18
feat: Recovery tests for orchestrator and worker crashes

### LINKED ISSUE #590: Integration test for journalling / recovery
Once we have real journalling (NATS?) and lease management (etcd) we should setup a suite of "real" integration tests that take chaos-monkey style approaches -- crashing workers and orchestrator and ensuring that flows eventually complete.

bjchambers:
## Summary

This closes #590.

- Add Docker Compose-based integration tests that deploy 2 orchestrators with shared SQLite + etcd, verifying recovery across six failure scenarios
- Add configurable remote retry with fibonacci/exponential/constant backoff for HTTP workers (via `backon` crate)
- Add attempt tracking to verify step execution counts across crashes
- Fix 3 bugs in the recovery system discovered during integration testing
- Open #637 for attempt-count reset bug after crash recovery

## Recovery Test Scenarios

| Scenario | Test | Description |
|----------|------|-------------|
| **A: Restart** | `test_restart_recovery_*` | Orchestrator killed and restarted (same ID) — recovers its own orphaned runs |
| **B: Failover** | `test_failover_recovery_*` | Orchestrator killed and stays dead — survivor claims orphans after lease TTL |
| **C: Dual failure** | `test_dual_failure_recovery` | Both orchestrators killed and restarted — both recover without duplication |
| **D: Worker crash** | `test_worker_crash_single_retry` | Worker process crashes (SIGUSR1 → `os._exit`), Docker auto-restarts, orchestrator retries with fibonacci backoff |
| **E: Retries exhausted** | `test_worker_crash_exhausts_retries` | Worker stays down, all retry attempts fail, run ends in `failed` state |
| **F: Dual crash** | `test_worker_and_orchestrator_crash` | Both worker and orchestrator crash — after restart, journal replay + retry completes the run |

Each test verifies:
1. Flow was **still in-progress** when killed (pre-kill assertion)
2. Flow reaches expected terminal state (`completed` or `failed`)
3. Recovery **produced new work** (post-recovery assertion that final steps now exist)
4. **Completed steps aren't re-executed** (step1 tracker_attempt == 1)
5. **Attempt tracking** is correct (orchestrator attempt count + tracker-side execution count)

## Remote Retry with Configurable Backoff

Added retry support for remote HTTP workers via the `backon` crate. Previously only subprocess workers had retry logic. The retry config is per-plugin in `stepflow-config.yml`:

```yaml
plugins:
  worker:
    type: stepflow
    url: "http://worker:8080"
    retry:
      maxAttempts: 3
      backoff:
        type: fibonacci    # or: constant, exponential
        minDelayMs: 2000
        maxDelayMs: 10000
```

Backoff strategies: **fibonacci** (default, aggressive early retry), **exponential** (with configurable factor), **constant** (fixed delay). Jitter is always included.

## Attempt Reset Bug (#637)

The orchestrator's retry attempt counter resets to 1 after crash recovery because `JournalEvent` doesn't record attempt counts. Tests document this with comments referencing the issue.

### 1. SQLite migration race condition
Two orchestrators running migrations simultaneously hit `UNIQUE constraint` on migration records and `duplicate column` errors from concurrent `ALTER TABLE` statements.

**Fix**: `INSERT OR IGNORE INTO` for migration records + re-check-after-failure logic in `apply_migration`.

### 2. Push-mode orphan detection gap
etcd `watch_orphans()` DELETE notification arrives before the heartbeat loop marks runs as orphaned in the metadata store. The push-mode loop finds nothing and waits indefinitely.

**Fix**: Add periodic fallback timer to `run_push_mode` alongside push notifications.

### 3. Recovery re-recovering active runs
The periodic recovery loop would re-recover runs already being actively executed, spawning duplicate `FlowExecutor`s.

**Fix**: Filter `runs_to_recover` against `active_executions.contains()` before spawning recovery.

## Test Infrastructure

```
tests/recovery/
├── docker-compose.yml          # etcd + 2 orchestrators + 1 worker (restart: unless-stopped)
├── Dockerfile.worker           # Python delay component image
├── stepflow-config.yml         # Shared config (etcd, sqlite, routes, retry)
├── components/delay_server.py  # Delay component with attempt tracking + SIGUSR1 crash handler
├── workflows/                  # Sequential (3 steps) and parallel (4+1) flows
├── conftest.py                 # Pytest fixtures (compose up/down with retry)
├── helpers.py                  # API client, docker control, crash_worker(), tracker reader
├── test_restart_recovery.py    # Scenario A (2 tests)
├── test_failover_recovery.py   # Scenario B (2 tests)
├── test_dual_failure.py        # Scenario C (1 test)
└── test_worker_crash.py        # Scenarios D, E, F (3 tests)
```

## Running

```bash
cd tests/recovery
uv run pytest -v --timeout=180
```

## Test plan

- [x] All Rust unit tests pass (700+ across workspace)
- [x] `./scripts/check-all.sh` passes (formatting, clippy, licenses, integration)
- [x] Recovery integration tests pass all 8 scenarios (requires Docker)

### DIFF
--- a/schemas/stepflow-config.json
+++ b/schemas/stepflow-config.json
@@ -483,81 +483,92 @@
     "BackoffConfig": {
       "oneOf": [
         {
-          "type": "object",
-          "description": "Fixed delay between retries.",
-          "required": [
-            "type"
-          ],
-          "properties": {
-            "delayMs": {
-              "type": "integer",
-              "format": "int64",
-              "description": "Delay in milliseconds (default: 1000).",
-              "minimum": 0
-            },
-            "type": {
-              "type": "string",
-              "const": "constant"
-            }
-          }
+          "$ref": "#/$defs/BackoffConfigConstant"
         },
         {
-          "type": "object",
-          "description": "Exponential backoff (delay doubles each attempt by default).",
-          "required": [
-            "type"
-          ],
-          "properties": {
-            "minDelayMs": {
-              "type": "integer",
-              "format": "int64",
-              "description": "Starting delay in milliseconds (default: 1000).",
-              "minimum": 0
-            },
-            "maxDelayMs": {
-              "type": "integer",
-              "format": "int64",
-              "description": "Maximum delay cap in milliseconds (default: 10000).",
-              "minimum": 0
-            },
-            "factor": {
-              "type": "number",
-              "format": "float",
-              "description": "Multiplier per attempt (default: 2.0)."
-            },
-            "type": {
-              "type": "string",
-              "const": "exponential"
-            }
-          }
+          "$ref": "#/$defs/BackoffConfigExponential"
         },
         {
-          "type": "object",
-          "description": "Fibonacci backoff (delay follows the Fibonacci sequence).",
-          "required": [
-            "type"
-          ],
-          "properties": {
-            "minDelayMs": {
-              "type": "integer",
-              "format": "int64",
-              "description": "Starting delay in milliseconds (default: 1000).",
-              "minimum": 0
-            },
-            "maxDelayMs": {
-              "type": "integer",
-              "format": "int64",
-              "description": "Maximum delay cap in milliseconds (default: 10000).",
-              "minimum": 0
-            },
-            "type": {
-              "type": "string",
-              "const": "fibonacci"
-            }
-          }
+          "$ref": "#/$defs/BackoffConfigFibonacci"
         }
       ],
-      "description": "Backoff strategy for retry delays. Each variant carries only its relevant parameters."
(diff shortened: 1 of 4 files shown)


schemas/stepflow-config.json
```diff
@@ -445,6 +480,85 @@
         }
       }
     },
+    "BackoffConfig": {
```
bjchambers: I believe this should have a discriminator as described here:

https://swagger.io/docs/specification/v3_0/data-models/inheritance-and-polymorphism/#discriminator

We may need to generate the schema manually.

PR #645, opened 2026-02-19
feat: Add TasksStarted journal event, simplify journal trait

### LINKED ISSUE #637: Retry attempt count resets after orchestrator crash recovery
## Bug

When an orchestrator crashes mid-retry and recovers, the retry attempt counter resets to 1. This means the total number of attempts across crashes can exceed `MAX_ATTEMPTS`.

## Root Cause

`JournalEvent` in `stepflow-rs/crates/stepflow-state/src/execution_journal.rs` has no event for recording attempt counts. The journal tracks:
- `RunCreated`, `RunInitialized`, `RunCompleted`
- `TaskCompleted` (with result)
- `StepsUnblocked`, `ItemCompleted`

When recovery replays the journal, it reconstructs the run state and re-dispatches in-flight steps. Since no attempt information is journaled, the retry loop in `StepflowPlugin::execute()` starts fresh at `attempt=1`.

## Impact

- A step that was on attempt 2 of 3 before a crash will get 3 more attempts after recovery (5 total instead of 3).
- In the worst case, repeated crashes could cause unlimited retries.

## Suggested Fix

Add a `TaskStarted` journal event:
```rust
TaskStarted {
    item_index: u32,
    step_index: usize,
    attempt: u32,
}
```

Record this event before each execution attempt. During recovery, replay these events to restore the attempt counter for in-flight steps.

## Note

This may mean that all journal events must be committed immediately. If so, should simplify the journal trait and implementations to use a "flush-immediately" style rather than the "write and then call flush" API it has now.

## Verification

The recovery integration tests in `tests/recovery/test_worker_crash.py` (scenario F: worker + orchestrator crash) demonstrate this behavior — the recovered orchestrator dispatches with `attempt=1` even though the step may have been attempted before the crash.

bjchambers:
## Summary

- Adds a `TasksStarted` journal event written before spawning task futures, recording execution-level attempt counts per task. On recovery replay, `TasksStarted` without a matching `TaskCompleted` indicates the task was in-flight during the crash, enabling accurate attempt tracking across orchestrator restarts.
- Simplifies the `ExecutionJournal` trait from `append` + `flush` (two-phase) to a single durable `write` method. All three implementations (InMemory, SQLite, NoOp) already committed writes immediately, making `flush` a no-op. The `TasksStarted` write naturally replaces the old `flush` barrier.
- Adds per-step attempt tracking in `ItemState` (`Vec<u32>` with `attempt_count()` / `record_attempt()` / `set_attempt_at_least()` methods).
- Passes the orchestrator-level attempt count through to component execution via `Plugin::execute()`.
- **Moves `run_id` from `JournalEntry` envelope into each `JournalEvent` variant**, eliminating the awkward per-`run_id` grouping of `TasksStarted` writes. Introduces `RunTaskAttempts` struct for multi-run `TasksStarted` batches.
- Simplifies the `write()` signature to `fn write(&self, root_run_id: Uuid, event: JournalEvent)` and `read_from()` to return `Vec<JournalEvent>`, hiding `JournalEntry` from the trait API.
- Adds `JournalEvent::involves_run()` helper for filtering events by run in shared journals.
- **Fixes execution tree recovery** (related to #651): Recovery was processing each run independently, creating broken `FlowExecutor`s for subflows with wrong `root_run_id`. Now groups discovered runs by `root_run_id` and only recovers the root run from each tree. `RunState::apply_event` filters events by `run_id` internally, so the full journal can be replayed without pre-filtering. Orphaned subflows without a running root are marked Failed.

**Attempt vs Retry**: The execution-level `attempt` counter (orchestrator level: "how many times have we tried to execute this step") is distinct from the transport-level `retry` counter in `plugin.rs` ("make sure the component receives the request"). Transport retries happen within a single execution attempt and do not change the attempt number visible to the component.

This is a prerequisite for future `ErrorAction::Retry` (step-level retry counting across crashes) and improves observability of in-flight tasks during crashes.

Closes #637

## Test plan

- [x] 4 unit tests in `run_state.rs`: single-task attempt increment, batch attempt tracking, mixed recovery scenario, compacted journal recovery (attempt > 1)
- [x] 2 integration tests in `recovery.rs`: end-to-end attempt=2 after recovery, batched `TasksStarted` for parallel tasks after recovery
- [x] 3 tests for execution tree recovery: root groups by `root_run_id`, orphaned subflows without root marked failed, root ignores interleaved subflow events
- [x] 1 test for cross-run event filtering in `run_state.rs`: events from other runs silently ignored
- [x] Journal compliance tests updated for new `write(root_run_id, event)` API and `RunTaskAttempts`
- [x] All existing tests updated for `append`→`write` rename, `run_id` in events, and `RunTaskAttempts`
- [x] `cargo test --workspace` — all tests pass
- [x] `cargo fmt --check` and `cargo clippy` clean

### DIFF
--- a/stepflow-rs/crates/stepflow-execution/src/step_runner.rs
+++ b/stepflow-rs/crates/stepflow-execution/src/step_runner.rs
@@ -238,7 +238,7 @@ pub struct StepRunner {
     step_input: FlowResult,
     /// Run context containing run hierarchy, environment, flow, and subflow submission.
     run_context: Arc<RunContext>,
-    /// Execution attempt number (1-based, increases across all retry causes).
+    /// Orchestrator-level attempt number (1-based, increases on journal recovery).
     attempt: u32,
 }
 
--- a/stepflow-rs/crates/stepflow-plugin/src/plugin.rs
+++ b/stepflow-rs/crates/stepflow-plugin/src/plugin.rs
@@ -48,8 +48,11 @@ pub trait Plugin: Send + Sync {
     /// * `run_context` - The run context with flow, environment, and subflow submission
     /// * `step` - The step being executed (None for workflow-level operations)
     /// * `input` - The resolved input values
-    /// * `attempt` - The execution attempt number (1-based). Increases on every retry
-    ///   regardless of cause (orchestrator crash, transport failure, etc.)
+    /// * `attempt` - The orchestrator-level execution attempt number (1-based).
+    ///   Increases when the orchestrator recovers a task from its journal.
+    ///   Implementations with transport-level retries (e.g., StepflowPlugin)
+    ///   may add the retry count to produce a combined attempt number for
+    ///   the component.
     async fn execute(
         &self,
         component: &Component,
--- a/stepflow-rs/crates/stepflow-protocol/src/plugin.rs
+++ b/stepflow-rs/crates/stepflow-protocol/src/plugin.rs
@@ -707,11 +707,13 @@ impl Plugin for StepflowPlugin {
             input
         };
 
-        // The component sees the orchestrator-level attempt number.
-        // Transport-level retries happen within a single execution attempt
-        // and do not change the attempt number visible to the component.
+        // The component sees a combined attempt number: the orchestrator-level
+        // attempt (from journal replay after crashes) plus transport-level retries
+        // (from connection failures within a single execution). This means the
+        // component always sees a monotonically increasing attempt number regardless
+        // of whether the retry was caused by an orchestrator crash or a worker crash.
         for retry in 0..max_retries {
-            let component_attempt = attempt;
+            let component_attempt = attempt + retry;
 
             // Create observability context from run context and step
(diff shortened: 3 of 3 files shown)


stepflow-rs/crates/stepflow-execution/src/flow_executor.rs
```diff
@@ -270,33 +270,35 @@ impl FlowExecutor {
...
-                        TaskAttempt {
-                            item_index: task.item_index,
-                            step_index: task.step_index,
-                            attempt: item.attempt_count(task.step_index),
-                        }
-                    })
-                    .collect();
-
-                // Write TasksStarted event — this must be durable before we spawn
+                // spawning. Tasks may belong to different runs (parent and subflows),
```
bjchambers: We shouldn't need to group things this way -- the journal is grouped by *root* run ID. I believe we should change the `write_event` API to *not* include the `run_id` and make that part of specific events. Thoughts?

PR #655, opened 2026-02-23
feat: Trait-based selective migration for store initialization

zznate:
## Summary

- Add `initialize_blob_store()`, `initialize_metadata_store()`, and `initialize_journal()` to `BlobStore`, `MetadataStore`, and `ExecutionJournal` traits with default no-op implementations
- `SqliteStateStore` implements each to run only the migrations needed for that store role
- Split monolithic `run_migrations()` into `run_blob_migrations()`, `run_metadata_migrations()`, `run_journal_migrations()`, and `run_all_migrations()`
- Drop FK constraint from `runs.flow_id` since blobs may live in a different backend (e.g., S3)
- `storage_config.rs` calls `initialize_*()` on each trait object after creation, so expanded-form configs only create the tables they need

## Test plan

- [x] Selective migration tests: blob-only, metadata-only, journal-only each create only their tables
- [x] Incremental migration test: adding metadata after blob works correctly
- [x] Idempotency and concurrent pool migration tests pass
- [x] All compliance tests (blob, metadata, journal) pass
- [x] `./scripts/check-rust.sh` passes (fmt, clippy, tests, docs)

### DIFF
--- a/stepflow-rs/crates/stepflow-config/src/storage_config.rs
+++ b/stepflow-rs/crates/stepflow-config/src/storage_config.rs
@@ -207,41 +207,57 @@ impl StorageConfig {
     ///
     /// When multiple stores have identical configurations, they share a single
     /// backend instance (smart deduplication based on full config equality).
+    /// Each store's `initialize_*` method is called to set up only the tables
+    /// needed for its role.
     pub async fn create_stores(&self) -> Result<Stores> {
         let (metadata_config, blobs_config, journal_config) = self.get_configs();
 
-        // Cache for deduplication: configs with the same values share one instance
+        // Deduplicate: create each unique config only once
         let mut cache: HashMap<StoreConfig, ConcreteStore> = HashMap::new();
-
-        // Helper to get or create a store for a given config
-        async fn get_or_create(
-            cache: &mut HashMap<StoreConfig, ConcreteStore>,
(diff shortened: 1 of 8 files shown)


stepflow-rs/crates/stepflow-config/src/storage_config.rs
```diff
@@ -75,12 +77,18 @@ impl ConcreteStore {
...
 
-/// Create a ConcreteStore from a StoreConfig.
-async fn create_concrete(config: &StoreConfig) -> Result<ConcreteStore> {
+/// Create a ConcreteStore from a StoreConfig with the given capabilities.
+///
+/// Capabilities are only meaningful for SQLite stores (controlling which tables
+/// get created). InMemory and Filesystem stores ignore the parameter.
+async fn create_concrete(
+    config: &StoreConfig,
+    capabilities: &StoreCapabilities,
```
bjchambers: I wonder if it would be simpler to just add an `initialize_metadata_store()`, `initialize_blob_store()` etc. to each of the traits. Then, when we create them, the configuration can just call the appropriate methods. Seems like less machinery than a separate `StoreCapability`.

And, it ensures that any *other* stores we create are properly initialized (eg., if we had PG, or C* it would ensure that it has separate initialization methods for each of the capabilities).

PR #658, opened 2026-02-23
feat: Subflow recovery (#651)

### LINKED ISSUE #651: Implement and verify recovery of subflows
Currently, I believe that recovery of subflows has issues.

The current code re-executes parent steps that were waiting for subflows. It doesn't resume running subflows in-place. 

Here's why this is hard:

1. A parent step waiting for a subflow is a running component function blocked on `wait_for_completion(subflow_run_id)`.
2. We can't resume a component mid-execution. After recovery, that step is "in-flight" (`TaskStarted`, no `TaskCompleted`), so the executor re-runs it.
3. The component generates a new `run_id` (via `Uuid::now_v7()`) and submits a new subflow, losing the old subflow's progress.

To achieve the desired behavior (resume subflows, parent steps pick up where they left off), we'd need:

1. A new journal event like `SubflowSubmitted { parent_run_id, parent_step_index, subflow_run_id }` to record which step is waiting for which subflow (or adding an optional parent-step index to the the existing event for launching for a subflow).
2. During recovery, recognize the parent step is "waiting for subflow" and skip re-executing the component. Instead, create a synthetic future that waits for the old subflow's `wait_for_completion`.
3. Recover the subflow's `RunState` and inject it into the `FlowExecutor` (needs additional_runs on `FlowExecutorBuilder`).
4. When the subflow completes, use its result as the parent step's result

We'd also likely need a way to handle *multiple* subflow submits from the same parent_step_index (eg., including a parent_step_subflow_count) or something and requiring subflows be submitted with deterministic order (or allowing a "subflow key" to be specified and defaulting to the subflow count.

# Tasks
- [ ] Add or extend the subflow event to support associating with specific steps.
- [ ] Add or extend the subflow API so the component can provide a key to use for identifying the submission (default to a sequence number within the component)
- [ ] Update recovery to restart subflows
- [ ] Update subflow execution to check for a recovered subflow and return it's run ID if available.
- [ ] We should add recovery integration tests that verify that recovery while one or more subflows are in flight resumes them rather than starting new subflows or leaving them dangling. We should also make sure that their completion properly progresses the outer flow. And that completed subflows don't get restarted.

# Further complications (likely separate issue)

Some components (like the looping components) will likely need a way to interact with journalling. Specifically, we probably *don't* want to have to iterate over all iterations to start the next iteration. Instead, we likely want a way to allow the component to record information about what iteration it is on and read that back (eg., allowing component level journalling and recovery). This would *only* be for components that are particularly stateful and don't want to rely on "replay everything I did and just get the results faster" (which would work in many cases).

bjchambers:
## Summary

Implements subflow recovery so that when an orchestrator crashes while a parent step is waiting for a subflow, recovery resumes the subflow in-place rather than restarting it from scratch. Closes #651.

### Commit 1: Add subflow key plumbing for recovery dedup
- Thread a deterministic `subflow_key` (UUID v5) through the entire submission path: Rust `SubflowSubmitter`, `RunContext`, protocol, Python SDK, and `runs/submit` handler
- Auto-generate keys via `Uuid::new_v5(run_id, &counter)` ensuring identical keys on re-execution after recovery
- Change `SubmitRunHandler` to use in-process subflow channel when a `SubflowSubmitter` is available, making Python-submitted subflows true in-tree subflows

### Commit 2: Add SubflowSubmitted journal event
- New `SubflowSubmitted` journal variant recording `(parent_run_id, item_index, step_index, subflow_key) → subflow_run_id`
- Written after `RunCreated` in `handle_submit_request()` to enable recovery deduplication

### Commit 3: Recover subflows in-place during crash recovery
- Reconstruct subflow `RunState` objects from journal replay during `recover_execution_tree()`
- Build dedup map from `SubflowSubmitted` events; when parent steps re-execute and re-submit subflows with matching keys, return existing `run_id`
- Integration tests: restart recovery, failover recovery, completed-subflow-not-restarted

## Test plan
- [x] All 27 workspace test suites pass (`cargo test --workspace --exclude stepflow-state-etcd`)
- [x] Recovery integration tests (`tests/recovery/test_subflow_recovery.py`) via Docker Compose

### DIFF
--- a/tests/recovery/stepflow-config.yml
+++ b/tests/recovery/stepflow-config.yml
@@ -14,6 +14,8 @@ plugins:
 routes:
   "/test/{*component}":
     - plugin: test-worker
+  "/builtin/{*component}":
+    - plugin: builtin
   "/{*component}":
     - plugin: builtin
 
--- a/tests/recovery/test_subflow_recovery.py
+++ b/tests/recovery/test_subflow_recovery.py
@@ -34,6 +34,7 @@ import pytest
 from helpers import (
     ORCH1_URL,
     ORCH2_URL,
+    clear_tracker,
     count_step_executions,
     docker_kill,
     docker_start,
@@ -62,6 +63,9 @@ async def test_subflow_restart_recovery(compose_env):
     6. Assert: inner_delay executed exactly once (recovered, not restarted)
     7. Assert: final_step executed once
     """
+    # Clear tracker so we only see records from this test.
+    clear_tracker()
+
     flow_id = await store_flow(ORCH1_URL, str(WORKFLOWS / "subflow_delay.yaml"))
     run_id = await submit_run(ORCH1_URL, flow_id, {"data": {"value": "subflow_test"}})
 
@@ -92,11 +96,12 @@ async def test_subflow_restart_recovery(compose_env):
 
     # Key assertion: inner_delay should have executed exactly once.
     # Without subflow recovery, it would execute twice (once before crash, once after).
-    assert count_step_executions(records, "inner_delay", run_id) == 1, (
+    # Note: inner_delay runs in the subflow (different run_id), so don't filter by parent run_id.
+    assert count_step_executions(records, "inner_delay") == 1, (
         "inner_delay should execute exactly once — subflow recovery should prevent re-execution"
     )
 
-    # final_step should have executed once after recovery
+    # final_step runs in the parent flow so its run_id matches
     assert count_step_executions(records, "final_step", run_id) >= 1, (
         "final_step should have executed after recovery"
     )
@@ -114,6 +119,8 @@ async def test_subflow_failover_recovery(compose_env):
     5. Assert: inner_delay executed exactly once
     6. Assert: final_step executed once
     """
+    clear_tracker()
+
     flow_id = await store_flow(ORCH1_URL, str(WORKFLOWS / "subflow_delay.yaml"))
     run_id = await submit_run(ORCH1_URL, flow_id, {"data": {"value": "failover_test"}})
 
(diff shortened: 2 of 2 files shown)


stepflow-rs/crates/stepflow-plugin/src/context.rs
```diff
@@ -244,6 +254,14 @@ impl RunContext {
     ) -> crate::Result<Vec<FlowResult>> {
         use error_stack::ResultExt as _;
 
+        // Generate a deterministic subflow key from the run_id and a monotonic counter.
```
bjchambers: This is potentially problematic if the client side also (intermittently) generates IDs using UUID v5 and the run ID and counter. Specifically,

1. Client generates UUID v5 with (run_id, counter=0) and sends it
2. Client submits subflow *without* counter. Orchestrator generates UUID v5 with (run_id, counter=0).

This leads to a collision. We probably want something in the execution state that records the UUIDs subflows have been created with so that (a) this can increment the counter until it finds a non-colliding UUID and (b) client-facing keys can be verified as unique.

stepflow-rs/crates/stepflow-protocol/src/protocol/flows.rs
```diff
@@ -43,6 +44,14 @@ pub struct SubmitRunProtocolParams {
     /// Observability context for tracing.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub observability: Option<ObservabilityContext>,
+    /// Caller-provided key for subflow deduplication during recovery.
+    ///
+    /// When a component submits a subflow via the bidirectional protocol,
+    /// it can provide a key that uniquely identifies this submission within
+    /// the scope of `(parent_run_id, item_index, step_index)`. If not
+    /// provided, a random UUID is used (no recovery deduplication).
```
bjchambers: This isn't true though -- the UUID may be generated sequentially and some recovery may happen. Need to make sure this is properly documented.

stepflow-rs/crates/stepflow-execution/src/recovery.rs
```diff
@@ -514,11 +520,145 @@ async fn recover_execution_tree(
...
+            Some(f) => f,
+            None => {
+                log::warn!(
+                    "Subflow flow not found for run {}, flow_id={}, skipping",
+                    sub_run_id,
+                    sub_flow_id
+                );
+                continue;
+            }
+        };
```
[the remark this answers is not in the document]
bjchambers: Fixed in ee95be2e. Recovery now fails explicitly with an error message when a subflow's flow definition is missing from the blob store, exactly as suggested. This prevents the dangling subflow_map entry that would cause wait_for_completion to hang.

PR #678, opened 2026-02-24
feat: Add docling step worker integration

### LINKED ISSUE #676: Build a drop-in replacement for the docling-serve sidecar that exposes the same REST API while calling docling's `DocumentConverter` library directly in-process
Currently, every docling-worker pod runs two processes: the Stepflow component server (Python) and a docling-serve sidecar (Python + FastAPI + Uvicorn). The component server is a thin HTTP proxy — it receives Stepflow JSON-RPC requests, reformats them, and forwards to the sidecar on localhost:5001. 

This works fine for translating langflow-docling steps as they come in, but could be made more efficient by calling the docling library directly. In other words, we are translating the interop of docling-serve <--> docling into native Stepflow. 

See the design document for this phase here: https://github.com/stepflow-ai/stepflow/blob/main/docs/proposals/docling-step-worker.md

zznate:
Add integrations/docling-step-worker/, a Stepflow component server that wraps docling's DocumentConverter library directly, eliminating the docling-serve HTTP sidecar.

Components:
- /docling/classify: lightweight PDF probing via pypdfium2
- /docling/convert: in-process DocumentConverter.convert()
- /docling/chunk: HybridChunker for semantic chunking

Key design:
- One DocumentConverter per named pipeline config (lazy init)
- DocumentStream for in-memory processing (no temp files)
- asyncio.to_thread() for sync docling calls
- Binary documents flow through Stepflow blob store

Also includes flow definitions, Dockerfile, CLI entry points, 59 unit tests, and integration parity tests.

Closes #676


bjchambers (commented):
I don't think the flows actually work, and I don't see any tests for running the flows. We should have a test htat uses stepflow-orchestrator to startup a local orchestrator and then submits the flows.

PR #692, opened 2026-02-25
feat: Unify retry counters into orchestrator with separate budgets

bjchambers:
## Summary

Moves all retry logic from `StepflowPlugin` into `FlowExecutor` so every attempt is journaled and the component sees a single monotonically increasing `attempt` counter regardless of retry reason.

- **Three retry reasons with separate budgets**: transport errors (orchestrator `retry.transportMaxRetries`), component errors (step-level `onError: { action: retry, maxRetries }`), and orchestrator recovery (unlimited)
- **Single monotonic attempt counter**: component always sees `attempt=1, 2, 3, ...` regardless of which reason triggered the retry
- **Journal durability**: every retry is journaled as `TasksStarted` before execution, so recovery always produces the correct next attempt number
- **Configurable shared backoff**: one backoff config at orchestrator level (constant, exponential, or fibonacci) shared across all retry types
- **OTel metrics**: `step.retries_total`, `step.retries_exhausted_total`, `step.executions_total` with reason/component labels

### Changes by area

**Unified error codes** (`stepflow-core`):
- New `ErrorCode` struct as single source of truth for all error codes across Rust, Python, and protocol schemas
- Range-based classification: JSON-RPC standard (-32700 to -32600), Stepflow protocol (-32099 to -32000), orchestrator/component (1–4999), transport (5000+)
- `JsonSchema` impl with `x-enum-varnames` auto-generates matching `ErrorCode(IntEnum)` in Python SDK
- Transport error detection changed from `code == 503` to `code >= 5000`
- Replaced all magic error code numbers with `ErrorCode::*` constants
- Fixed swapped `invalid_value`/`not_found` codes in protocol `Error` helpers

**Component vs transport error distinction** (`stepflow-protocol`):
- `StepflowPlugin::execute()` now converts `MethodError` responses to `FlowResult::Failed` with the original error code, instead of treating all plugin errors as transport failures
- Only true transport errors (subprocess crash, network timeout) produce `Err(...)` → `TRANSPORT_ERROR (5000)`
- Component server errors (e.g. `-32004 COMPONENT_EXECUTION_FAILED`) are now `Ok(FlowResult::Failed(...))` with their original code preserved

**Core types** (`stepflow-core`):
- `ErrorAction::Retry` gains `maxRetries: Option<u32>` (default 3)
- New `RetryConfig` with `transportMaxRetries` and configurable `BackoffConfig` (constant/exponential/fibonacci)
- `FlowResult::is_transport_error()` uses `ErrorCode::is_transport(code)` (range-based)

**Plugin trait** (`stepflow-plugin`):
- New method: `prepare_for_retry()` — plugins recover resources (e.g. restart subprocess) before retry
- `transport_max_retries()` removed — retry limits are now orchestrator-level, not per-plugin

**StepflowPlugin** (`stepflow-protocol`):
- Retry loop removed — single attempt per `execute()` call
- Per-plugin `RetryConfig`/`BackoffConfig` removed — moved to orchestrator
- `backon` dependency removed

**Environment** (`stepflow-plugin`):
- `StepflowEnvironmentBuilder` gains `.retry()` setter
- `RetryConfig` stored in environment, read by executor

**Execution engine** (`stepflow-execution`):
- `RetryState` moved from `FlowExecutor`'s `HashMap` into `ItemState` — eliminates manual eviction
- `FlowExecutor` gains `maybe_retry_task()` and configurable backoff
- Transport error detection via `ErrorCode::is_transport()` (code >= 5000)
- `ErrorAction::Retry` implemented for component errors

**Observability** (`stepflow-observability`):
- Three new step retry metrics with reason/component labels

**Config** (`stepflow-config`):
- New top-level `retry` field (replaces per-plugin retry config)

**Python SDK**:
- `ErrorCode` auto-generated from protocol schema (replaces hand-maintained enum with wrong names)
- `exceptions.py` imports from generated code, fixes all default error codes
- `http_server.py` uses `ErrorCode.*` constants instead of magic numbers
- `generate.py` adds `--use-subclass-enum` for integer enum generation

**Documentation**:
- `errors.md` updated with all named error codes and ranges
- Transport error code changed from 503 to 5000+

## Test plan
- [x] All Rust tests pass (`cargo test`)
- [x] Python linter passes (`ruff check`)
- [x] Schema regenerated (protocol.json with `ErrorCode` in `$defs`)
- [x] Python codegen regenerated (`ErrorCode(IntEnum)` with all variants)
- [x] Mock test expectations updated for transport error code (5000)
- [x] Error handling test expectations updated for component error code (-32004)
- [ ] Recovery integration test (requires Docker/k8s environment)

### DIFF
--- a/stepflow-rs/crates/stepflow-observability/src/metrics.rs
+++ b/stepflow-rs/crates/stepflow-observability/src/metrics.rs
@@ -151,15 +151,6 @@ static STEP_EXECUTIONS: LazyLock<Counter<u64>> = LazyLock::new(|| {
         .build()
 });
 
-/// Aggregate step errors across all components (low-cardinality).
-static STEP_ERRORS: LazyLock<Counter<u64>> = LazyLock::new(|| {
-    let meter = global::meter("stepflow");
-    meter
-        .u64_counter("step.errors_total")
-        .with_description("Total step execution errors across all components")
-        .build()
-});
-
 /// Record a step retry.
 pub fn record_step_retry(reason: &str, component: &str) {
     STEP_RETRIES.add(
@@ -191,7 +182,4 @@ pub fn record_step_execution(component: &str, outcome: &str) {
             KeyValue::new("outcome", outcome.to_string()),
         ],
     );
-    if outcome == "failed" {
-        STEP_ERRORS.add(1, &[]);
-    }
 }


stepflow-rs/crates/stepflow-execution/src/flow_executor.rs
```diff
@@ -38,7 +38,33 @@ use crate::state::ItemsState;
...
+    /// Number of retries consumed due to transport errors (subprocess crash, network failure).
+    transport_retries: u32,
+    /// Number of retries consumed due to component errors (component returned an error).
+    component_retries: u32,
+}
+
+/// Compute fibonacci backoff delay for a retry.
+///
+/// Uses fibonacci sequence: 1s, 1s, 2s, 3s, 5s, 8s, 10s (capped).
+fn compute_retry_delay(retry_count: u32) -> std::time::Duration {
```
bjchambers: This should be configurable and use the backoff configurations we use elsewhere. This will allow the user to control how the retry backoff is computed.

stepflow-rs/crates/stepflow-execution/src/step_runner.rs
```diff
@@ -78,6 +78,11 @@ pub struct StepRunResult {
     pub metadata: StepMetadata,
     /// The execution result (success or failure).
     pub result: FlowResult,
+    /// True if the failure was caused by a transport/infrastructure error
+    /// (subprocess crash, network timeout, connection refused) rather than
+    /// a component logic error. Only meaningful when `result` is `Failed`.
+    #[serde(skip)]
+    pub is_transport_error: bool,
```
bjchambers: Rather than the somewhat arbitrary `is_transport_error` boolean flag, would it make sense to have one or more error codes dedicated to "transport errors" and use that? The `FlowResult` already carries an error code in appropriate scenarios.

stepflow-rs/crates/stepflow-plugin/src/plugin.rs
```diff
@@ -61,6 +60,20 @@ pub trait Plugin: Send + Sync {
...
+    /// timeout, connection refused) — cases where the component never ran or
+    /// didn't complete. Component logic errors are retried separately via
+    /// the step's `ErrorAction::Retry` and do not count against this budget.
+    fn transport_max_retries(&self) -> u32;
+
+    /// Prepare the plugin for a retry after a transport error.
+    ///
+    /// For subprocess plugins, this restarts the crashed process.
+    /// For remote plugins, this is a no-op (external orchestration handles restart).
+    async fn prepare_for_retry(&self) -> Result<()>;
```
bjchambers: It seems like this should default to a no-op implementation. Reword details to be less specific to two plugins since we may introduce more.

sdks/python/stepflow-py/src/stepflow_py/worker/exceptions.py
```diff
@@ -112,7 +106,7 @@ def __init__(self):
 
     @property
     def default_code(self) -> ErrorCode:
-        return ErrorCode.NOT_INITIALIZED
+        return ErrorCode.SERVER_NOT_INITIALIZED
```
zznate: Same here: `COMPONENT_SERVER_NOT_INITIALIZED`

stepflow-rs/crates/stepflow-builtins/src/iterate.rs
```diff
@@ -148,7 +148,7 @@ impl BuiltinComponent for IterateComponent {
 
                     // Result doesn't have expected structure - return error
                     return Ok(FlowResult::Failed(stepflow_core::FlowError::new(
-                        400,
+                        stepflow_core::ErrorCode::BAD_REQUEST,
```
zznate: The orchestrator, we can probably leave prefixes off of, so this is fine.

PR #706, opened 2026-03-03
docs: Stepflow 699 docling deploy blog

zznate:


docs/blog/2026-03-docling-step-worker-deploy.md
```diff
@@ -0,0 +1,332 @@
...
+
+Stepflow is a general purpose AI workflow system designed to solve exactly this class of problem. Its orchestration architecture provides resilient distributed execution state, HTTP SSE-aware load-balanced routing to worker pools, persistent result storage, and full observability out of the box. After initial success with scaling Langflow throughput, we started looking for other high-value integration projects. After looking through docling throughput as part of some day-job work with [OpenRAG](https://www.openr.ag/), we started to wonder: *how hard would it be to take a sophisticated AI pipeline like docling and run it entirely on Stepflow?*
```
bjchambers: I would start a new paragraph for the "After initial success ...". Basically, first paragraph is background on Docling. Second paragraph is background on Stepflow. Then we introducue the article.

Or, we could do "Stepflow is ...", then "After initial success ...", *then* docling (introduce the topic, then provide context).

Either way, split this paragraph.

docs/blog/2026-03-docling-step-worker-deploy.md
```diff
@@ -0,0 +1,332 @@
...
+
+**Built-in observability.** Every component execution emits OpenTelemetry traces, metrics, and logs. The production deployment includes a full o11y stack: Jaeger for traces, Prometheus for metrics, Loki for logs, and Grafana dashboards.
+
+**Recovery.** In-flight subflows are resumed during recovery. Execution checkpoints allow faster restart after failures.
+
+These aren't docling-specific features. They're what Stepflow provides to *any* pipeline that plugs into it. Running Docling this way just happens to be a great example of why they matter!
+
+## The Integration Architecture
+
+The example in this post showcases a simple setup of three worker pods running docling. While lightweight enough to run on a laptop, nothing is stopping you from adding instances, turning up the CPU and scaling this out. That said, our initial deployment runs in the `stepflow-docling` namespace with three pods:
```
bjchambers: "three worker pods running docling" -- this makes it sound like we *are* running docling serve.

I this should say:

In this post, we create a docling worker implementing the Stepflow protocol to provide the *steps* of the docling pipeline as Stepflow components. We run three instances of this service. We also create a Stepflow flow describing the structure of the docling pipeline using these components.

...

Eg., describe what *code* we wrote before discusing deployment so it is clear we're running that new code and how it is used.

docs/blog/2026-03-docling-step-worker-deploy.md
```diff
@@ -0,0 +1,332 @@
...
+  - id: chunk
+    component: /docling/chunk
+    input:
+      document: { $step: "convert", path: "document_dict" }
+      chunk_options: { $input: "chunk_options" }
+```
+
+This workflow uses three steps: classify, convert, chunk. Each step is a component that Stepflow routes to available workers through the load balancer. The flow is registered once at startup, then every document submission reuses the same flow definition with different inputs.
+
+This is the same pattern that makes Stepflow's Langflow integration work. Define the pipeline declaratively and let the orchestrator handle execution. The difference is that here we're integrating a purpose-built document processing library rather than a visual workflow builder, but the integration surface is basically the same.
```
bjchambers: Unrelated: This makes me think we should have a special kind of component that just provides a re-usable subflow definition. So we could say "this Langflow component expands to docling subflow". Could be a nice way to compose flows.

PR #714, opened 2026-03-04
feat: Docling facade parity with stock docling-serve

zznate:
Fix three bugs blocking the stepflow-docling facade from producing identical output to stock docling-serve, and add a batch comparison script that verifies parity across 10 ML papers.

Bug fixes:
- Resolve $blob references in worker: the orchestrator externalizes large base64 inputs into blob store, so the worker must detect {"$blob": "<sha256>"} dicts and fetch the original value before processing (blob_utils.py)
- Propagate blob API URL via STEPFLOW_BLOB_API_URL env var instead of relying on the initialize message (which the LB may proxy to a different worker). The SDK configures the blob store per-request within the correct async scope (server.py, http_server.py)
- Raise Axum body size limit to 250 MiB for large PDF uploads (startup.rs)

Parity fix:
- Default images_scale to 2.0 in the facade translator, matching stock docling-serve. Without this, embedded images render at ~0.4x the expected resolution.

Test tooling:
- Add compare-batch.py: downloads 10 arxiv papers and converts each through both the facade and stock docling-serve, reporting md similarity, text-only similarity, image counts, and wall times.

Results: 10/10 papers convert successfully with md similarity 1.00 and text similarity 0.99 average across the corpus.

### DIFF
--- a/stepflow-rs/crates/stepflow-server/src/startup.rs
+++ b/stepflow-rs/crates/stepflow-server/src/startup.rs
@@ -10,8 +10,8 @@
 // or implied. See the License for the specific language governing permissions and limitations under
 // the License.
 
-use axum::extract::DefaultBodyLimit;
 use axum::Router;
+use axum::extract::DefaultBodyLimit;
 use error_stack::ResultExt as _;
 use std::sync::Arc;
 use stepflow_config::{ConfigError, StepflowConfig};


bjchambers:
> Resolve $blob references in worker: the orchestrator externalizes large base64 inputs into blob store, so the worker must detect {"$blob": ""} dicts and fetch the original value before processing (blob_utils.py)

I don't think the orchestrator is doing the externalization, it's the Python worker SDK. But I would have expected it to *also* fetch the blob and turn it into bytes. This may be worth opening a bug for to investigate, rather than hacking it in the docling worker.

PR #727, opened 2026-03-06
docs: Initial proposal for backpressure focused on the docling-step-w…

zznate:
…orker use case

### DIFF
--- a/docs/proposals/backpressure.md
+++ b/docs/proposals/backpressure.md
@@ -1,0 +1,357 @@
+# Proposal: Backpressure for Stepflow Worker Pipeline
+
+**Status:** Draft
+**Authors:** Nate McCall
+**Created:** March 2026
+**Related:** `docling-step-worker.md` (facade architecture), Issue [#717](https://github.com/stepflow-ai/stepflow/issues/717) (custom metrics)
+
+## Summary
+
+This proposal adds backpressure to the Stepflow worker pipeline via two coordinated mechanisms: a **concurrency semaphore** in the Python SDK worker (Layer 1) that makes busy workers reject cleanly rather than fail silently, and a **dispatch semaphore** in the Rust orchestrator (Layer 2) that queues accepted runs rather than dispatching all of them immediately. Together they replace the current failure mode — transport errors, exhausted retries, cascading failures — with graceful queuing. Runs are durably accepted the moment they arrive. They wait for worker capacity to become available, then execute. Clients never see a new error shape.
+
+A `stepflow_pending_runs` Prometheus gauge, emitted as a natural byproduct of Layer 2, provides the autoscale signal for Kubernetes HPA/KEDA. New worker pods join the pool immediately and begin draining the queue via the existing least-connections routing.
+
+**Current state:** Unbounded. All requests accepted and dispatched immediately. Workers saturate. Transport retries (3 attempts, ~4s total) exhaust before a 25-second conversion finishes. Runs fail.
+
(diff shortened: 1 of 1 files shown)


docs/proposals/backpressure.md
```diff
@@ -0,0 +1,357 @@
...
+
+A `stepflow_pending_runs` Prometheus gauge, emitted as a natural byproduct of Layer 2, provides the autoscale signal for Kubernetes HPA/KEDA. New worker pods join the pool immediately and begin draining the queue via the existing least-connections routing.
+
+**Current state:** Unbounded. All requests accepted and dispatched immediately. Workers saturate. Transport retries (3 attempts, ~4s total) exhaust before a 25-second conversion finishes. Runs fail.
+
+**Target state:** Workers reject when busy (clean 503 + JSON-RPC error, retried by orchestrator). Orchestrator queues runs that can't dispatch immediately. Pending run depth signals autoscaling. New workers drain the queue automatically. Docling clients see no new error shapes.
```
bjchambers: I don't know that this is feasible. A *run* may have many different steps. Which workers should be consulted? What if during execution that run doesn't actually need the corresponding steps? Why did it queue? The needed workers are only known dynamically (during execution)

docs/proposals/backpressure.md
```diff
@@ -0,0 +1,357 @@
+# Proposal: Backpressure for Stepflow Worker Pipeline
+
+**Status:** Draft
+**Authors:** Nate McCall
+**Created:** March 2026
+**Related:** `docling-step-worker.md` (facade architecture), Issue [#717](https://github.com/stepflow-ai/stepflow/issues/717) (custom metrics)
+
+## Summary
+
+This proposal adds backpressure to the Stepflow worker pipeline via two coordinated mechanisms: a **concurrency semaphore** in the Python SDK worker (Layer 1) that makes busy workers reject cleanly rather than fail silently, and a **dispatch semaphore** in the Rust orchestrator (Layer 2) that queues accepted runs rather than dispatching all of them immediately. Together they replace the current failure mode — transport errors, exhausted retries, cascading failures — with graceful queuing. Runs are durably accepted the moment they arrive. They wait for worker capacity to become available, then execute. Clients never see a new error shape.
```
bjchambers: I don't think we should modify the orchestrator. If it is running something and the worker doesn't respond immediately that will slow down the execution (throttle the progress) automatically.

So, I think the worker semaphore shouldn't cause failure but throttling.

Then we can have the load balancer and/or routing layer do more to direct work to less loaded workers.

docs/proposals/backpressure.md
```diff
@@ -0,0 +1,357 @@
...
+Under concurrent load (concurrency 3, 10 PDFs), with less than 3 workers, all conversions fail:
+
+```
+[facade] [1/10] attention-is-all-you-need.pdf  FAIL  ({"detail": "step \"convert\" failed"...})
+[facade] [2/10] bert.pdf                        FAIL  ({"detail": "step \"convert\" failed"...})
+...
+[facade] Batch complete: 0/10 succeeded, total wall time: 308.6s
+```
+
+Root cause: each docling conversion takes 20-30 seconds of CPU-intensive work. With concurrency 3, the facade submits 3 flow runs simultaneously. The orchestrator dispatches all component executions immediately — there is no gating. Workers accept every request but can only process one conversion at a time. Requests to busy workers fail at the HTTP transport level. The orchestrator retries 3 times with fibonacci backoff (1s + 1s + 2s = 4s total). A worker blocked for 25 seconds will still be blocked when all 3 retries have been spent. The run fails. There are settings within the docling runtime to increase parallelism (more details below), but this will impact throughput. We are also leaving these as is due to it presenting a real-world issue that, given our architecture, we should be able to easily handle.
```
bjchambers: Why does the worker fail at the http level? It should just respond later.

Issue #716, opened 2026-03-04
Blob store initialization and $blob resolution broken behind load balancer


zznate:
Blob store: LB does not forward initialization to workers, breaking $blob resolution and blob API URL propagation. In the context of working #708 we came across some issues with blob store that required a workaround as discussed in the following PR: https://github.com/stepflow-ai/stepflow/pull/714

Basically, when a component server runs behind the load balancer, blob store functionality is broken because the `initialize` message (which carries `RuntimeCapabilities` including `blobApiUrl` and `blobThreshold`) is sent once by the orchestrator and routed by the LB to a single worker. All other workers behind the LB never receive initialization, leaving their blob configuration unset.
 
This causes two failures:
 
#### 1. `$blob` references are not resolved
 
The orchestrator blobifies large input fields in `StepflowPlugin::execute()` (`plugin.rs:443-458`), replacing them with `{"$blob": "<sha256>", "blobType": "data", "size": N}` dicts. On the SDK side, `server.py:561-566` has `resolve_blob_refs()` that should auto-resolve these back to inline data before passing to components — but it only runs when `_blob_threshold > 0`:
 
```python
# server.py:561-566
raw_input = params.input
if self._blob_threshold > 0:
    raw_input = await resolve_blob_refs(raw_input)
```
 
`_blob_threshold` is only set during `_handle_initialize()` (`server.py:399-409`) when both `blobThreshold` and `blob_api_url` are received from the orchestrator. Workers that never received `initialize` have `_blob_threshold == 0` and skip resolution entirely. Components receive raw `$blob` dicts instead of the actual data.
 
#### 2. Blob API URL is not available for worker-initiated blob operations
 
Workers that need to store or retrieve blobs (via `StepflowContext.put_blob()` / `get_blob()` or the module-level `blob_store` API) require `blob_api_url` to be configured. This URL is only propagated during initialization. Uninitialized workers have `_blob_api_url = None` and any blob operation fails with "Blob API URL not configured".
 
### Root Cause
 
The load balancer is a transparent HTTP proxy. It forwards the single `initialize` request to whichever worker the least-connections algorithm selects. It does not:
- Capture the initialization parameters
- Replay initialization to other existing workers
- Initialize new workers that join after scale-up
 
### Current Workaround (PR #714)
 
The docling-step-worker works around both issues:
 
[...]

PR #734, opened 2026-03-10
feat: Add Protocol Buffers + gRPC pull-based transport

bjchambers:
## Summary

Adds gRPC as an alternate transport layer between the orchestrator and workers, using a **pull-based task dispatch model**. Workers pull tasks from the orchestrator via a long-lived gRPC stream, execute them, and report results back. Each task carries self-describing `ExecutionContext` with callback URLs, eliminating the stateful initialization handshake (#716) and enabling natural backpressure.

The existing HTTP JSON-RPC transport is unchanged — this PR adds pull-based dispatch alongside it as a new `type: pull` plugin option.

### What's included

**Proto definitions** (7 files in `proto/stepflow/v1/`):
- Service definitions for flows, runs, blobs, components, tasks, orchestrator, health
- `google.api.http` annotations for REST transcoding via tonic-rest
- Proto-generated OpenAPI spec (`schemas/openapi.yaml`)

**Rust `stepflow-grpc` crate**:
- All gRPC service implementations (tonic) + REST routes at `/proto/api/v1`
- Self-contained in-memory work queue via `TasksService.PullTasks`
- `TaskCompletionRegistry` with two-phase lifecycle (Queued → Executing):
  - Queue timeout (configurable, default 30s) for dispatch-to-StartTask
  - Heartbeat crash detection (5s timeout, 1s check interval)
  - Optional execution timeout
- `PullPluginConfig` for `type: pull` plugins with subprocess management
- Structured gRPC errors via `tonic-types`
- 6 task lifecycle metrics (queue/execution duration, timeout/success/failure)

**Python gRPC worker**:
- Unified entry point — same `stepflow_py` binary for HTTP and gRPC modes
- Transport configured via `STEPFLOW_TRANSPORT` env var (default: `grpc`)
- Task service URL via `STEPFLOW_TASK_URL` env var (for K8s deployments)
- Background heartbeat loop, shared gRPC channel reuse

**CI**: proto lint, Python stub freshness checks, OpenAPI validation

### Architecture

```
Orchestrator                              Worker (stepflow_py)
┌─────────────────────┐                   ┌──────────────────┐
│ StepflowQueuePlugin │                   │ grpc_worker.py   │
│                     │◄──PullTasks───────│                  │
│ InMemoryTask        │   (gRPC stream)   │                  │
│   Transport         │                   │  StartTask ──────│──┐
│                     │                   │  Heartbeat ──────│──│ shared channel
│ TaskCompletion      │◄──CompleteTask────│  CompleteTask ───│──┘
│   Registry          │   (gRPC unary)    │                  │
└─────────────────────┘                   └──────────────────┘
```

### Next steps

This PR uses a **self-contained in-memory work queue** via `InMemoryTaskTransport`. The `TaskTransport` trait is designed to support pluggable backends. Future work will:

1. **Add external queue transports** — `TaskTransport` implementations for Kafka, NATS, MQ, etc., enabling distributed task dispatch across orchestrator instances. These will all use the same `type: pull` plugin with different transport configurations.
2. **Swap REST routes** — replace aide-generated `/api/v1` routes with proto-generated tonic-rest routes
3. **Deprecate and replace JSON-RPC** — once pull transport is validated, remove the HTTP JSON-RPC transport, SSE mechanism, and Pingora load balancer

See [docs/proposals/protobuf-grpc-migration.md](docs/proposals/protobuf-grpc-migration.md) for the full design.

## Test plan

- [x] `cargo check` — full workspace compiles clean
- [x] `cargo test -p stepflow-grpc` — 21 tests pass (including deterministic timeout tests with `tokio::time::pause()`)
- [x] `scripts/test-integration.sh tests/grpc/` — end-to-end tests (echo, error handling, multi-step, blob roundtrip, sub-run)

### DIFF
--- a/stepflow-rs/crates/stepflow-grpc/src/grpc_server.rs
+++ b/stepflow-rs/crates/stepflow-grpc/src/grpc_server.rs
@@ -15,9 +15,9 @@
 //! [`StepflowGrpcServer`] is created once by the orchestrator and hosts all
 //! worker-facing gRPC services:
 //!
-//! - [`TasksService`] — streams task assignments to workers (multi-queue)
-//! - [`OrchestratorService`] — handles task completion, sub-runs, heartbeats
-//! - [`BlobService`] — blob storage operations
+//! - `TasksService` — streams task assignments to workers (multi-queue)
+//! - `OrchestratorService` — handles task completion, sub-runs, heartbeats
+//! - `BlobService` — blob storage operations
 //!
 //! The server is stored in the [`StepflowEnvironment`] and started lazily
 //! on first use. Plugins register their task queues via [`register_queue`].
--- a/stepflow-rs/crates/stepflow-grpc/src/services/blob_service.rs
+++ b/stepflow-rs/crates/stepflow-grpc/src/services/blob_service.rs
@@ -23,7 +23,7 @@ use crate::proto::stepflow::v1::{
     GetBlobRequest, GetBlobResponse, PutBlobRequest, PutBlobResponse,
 };
 
-/// gRPC implementation of [BlobService].
+/// gRPC implementation of `BlobService`.
 #[derive(Debug)]
 pub struct BlobServiceImpl {
     env: Arc<StepflowEnvironment>,
--- a/stepflow-rs/crates/stepflow-grpc/src/services/orchestrator_service.rs
+++ b/stepflow-rs/crates/stepflow-grpc/src/services/orchestrator_service.rs
@@ -34,7 +34,7 @@ use crate::proto::stepflow::v1::{
     TaskHeartbeatResponse,
 };
 
-/// gRPC implementation of [OrchestratorService].
+/// gRPC implementation of `OrchestratorService`.
 ///
 /// This service is called by workers during component execution to submit
 /// sub-runs, query run status, and report task completions. Workers reach
--- a/stepflow-rs/crates/stepflow-grpc/src/services/tasks_service.rs
+++ b/stepflow-rs/crates/stepflow-grpc/src/services/tasks_service.rs
@@ -10,11 +10,11 @@
 // or implied. See the License for the specific language governing permissions and limitations under
 // the License.
 
-//! gRPC implementation of [`TasksService`].
+//! gRPC implementation of `TasksService`.
 //!
 //! This service runs on the orchestrator and is consumed by workers (workers
 //! are *clients* of this service). It streams task assignments from a shared
-//! [`PullTaskQueue`] to connected workers.
+//! [`PullTaskQueue`](crate::pull_task_queue::PullTaskQueue) to connected workers.
 //!
 //! Static worker configuration (blob service URL, blobification threshold,
 //! etc.) is provided to workers via environment variables at deployment time,
@@ -35,9 +35,9 @@ use crate::grpc_server::QueueRegistry;
 use crate::proto::stepflow::v1::tasks_service_server::TasksService;
 use crate::proto::stepflow::v1::{PullTasksRequest, TaskAssignment};
 
-/// gRPC implementation of [TasksService].
+/// gRPC implementation of `TasksService`.
 ///
-/// Routes `PullTasks` requests to the correct [`PullTaskQueue`] based on
+/// Routes `PullTasks` requests to the correct [`PullTaskQueue`](crate::pull_task_queue::PullTaskQueue) based on
 /// the worker's `queue_name`. Each pull plugin registers its queue under
 /// a unique name in the shared [`QueueRegistry`].
 #[derive(Debug)]


schemas/stepflow-config.json
```diff
@@ -419,6 +430,25 @@
...
+        "type": {
+          "type": "string",
+          "const": "grpc",
+          "default": "grpc"
+        }
+      },
+      "required": [
+        "url",
+        "type"
+      ]
```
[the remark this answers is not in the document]
bjchambers: Fixed in 4b4d8b6c — regenerated with `STEPFLOW_OVERWRITE_SCHEMA=1 cargo test -p stepflow-config test_schema_generation --all-features`. Now correctly shows `command`, `args`, `env`, `pluginName`, `queueTimeoutSecs`, `executionTimeoutSecs` with `type: pull` discriminator.

docs/proposals/protobuf-grpc-migration.md
```diff
@@ -0,0 +1,218 @@
...
+- REST route codegen from proto annotations
+- `prost-wkt-types` for serde-enabled WKTs (Value, Struct, etc.)
+
+### Phase 2: gRPC + REST server (complete)
+
+- All 6 gRPC service implementations on the orchestrator
+- REST routes mounted at `/proto/api/v1` alongside existing aide routes
+- Binary-aware blob routes with content negotiation
+- Proto-generated OpenAPI spec (`schemas/openapi-proto.json`)
+- Journal-based SSE streaming for `GetRunEvents`
```
zznate: Why SSE here?

docs/proposals/protobuf-grpc-migration.md
```diff
@@ -0,0 +1,218 @@
...
+| Proto file | Service | Transport | Purpose |
+|---|---|---|---|
+| `flows.proto` | `FlowsService` | REST + gRPC | Flow CRUD |
+| `runs.proto` | `RunsService` | REST + gRPC | Run lifecycle, SSE events |
+| `blobs.proto` | `BlobService` | REST + gRPC | Content-addressed blob storage |
+| `components.proto` | `ComponentsService` | REST + gRPC | Component listing |
+| `tasks.proto` | `TasksService` | gRPC only | Pull-based task dispatch |
+| `orchestrator.proto` | `OrchestratorService` | gRPC only | Worker callbacks (sub-runs, completion) |
+
+REST routes are auto-generated from `google.api.http` annotations using `tonic-rest-build`.
```
zznate: Maybe add a short bit pointing at `gnostic/openapi/v3` and `google/api` specifically.

proto/google/api/http.proto
```diff
@@ -0,0 +1,52 @@
...
+message Http {
+  repeated HttpRule rules = 1;
+  bool fully_decode_reserved_expansion = 2;
+}
+
+// HttpRule defines the mapping of an RPC method to one or more HTTP REST API methods.
+message HttpRule {
+  string selector = 1;
+
+  oneof pattern {
```
zznate: nit: do we want HEAD to do anything special eg. lightweight/no-body reporting on status?

proto/stepflow/v1/flows.proto
```diff
@@ -0,0 +1,218 @@
...
+  // The flow definition as JSON.
+  google.protobuf.Struct flow = 1;
+
+  // If true, only validate the flow without storing it.
+  bool dry_run = 2;
+}
+
+message StoreFlowResponse {
+  // Content-based flow ID (SHA-256 hash). Always present, even for dry runs.
+  string flow_id = 1;
```
zznate: Actually, if we did this as a type7 UUID, we would have timestamp for free.

zznate (commented):
Some questions and navel gazing.

proto/stepflow/v1/flows.proto
bjchambers: Should flow_id be modeled by a newtype or well-known type? Applies here and throughout.

proto/stepflow/v1/health.proto
```diff
@@ -0,0 +1,49 @@
...
+}
+
+message HealthCheckRequest {}
+
+message HealthCheckResponse {
+  // Service status (e.g., "healthy").
+  string status = 1;
+
+  // Timestamp when health was checked (RFC 3339 format).
+  string timestamp = 2;
```
bjchambers: This should use google.protobuf.Timestamp the well-known type. That will also provide better support for code-generation. We should do this for this timestamp and throughout.

proto/stepflow/v1/orchestrator.proto
```diff
@@ -0,0 +1,206 @@
...
+// --- Response ---
+
+message OrchestratorRunStatus {
+  string run_id = 1;
+  string flow_id = 2;
+  optional string flow_name = 3;
+  string status = 4;
+  ItemStatistics items = 5;
+  string created_at = 6;
+  optional string completed_at = 7;
```
bjchambers: timestamp

proto/stepflow/v1/runs.proto
```diff
@@ -0,0 +1,444 @@
...
+    option (gnostic.openapi.v3.operation) = {
+      summary: "Cancel a running execution"
+      operation_id: "cancelRun"
+    };
+  }
+}
+
+// --- CreateRun ---
+
+message CreateRunRequest {
```
bjchambers: It may make sense to share the CreateRunRequest between `runs.proto` (for creating root runs) and `orchestrator.proot` (used for creating sub-runs). They should have the same arguments.

sdks/python/stepflow-py/src/stepflow_py/worker/main.py
```diff
@@ -61,11 +66,61 @@ def main():
...
+        help="Use HTTP JSON-RPC transport",
+    )
+    parser.add_argument(
+        "--orchestrator-url",
+        type=str,
+        default=os.environ.get("STEPFLOW_TASK_URL", "localhost:7837"),
+        help="Task service gRPC address (env: STEPFLOW_TASK_URL)",
+    )
+    parser.add_argument(
+        "--plugin-name",
```
bjchambers: This is used to identify which queue to pull from. This should probably be `--queue-name` and `STEPFLOW_QUEUE_NAME` or something like that.

docs/proposals/protobuf-grpc-migration.md
bjchambers: The gRPC returns a `stream` which the gRPC *REST* layer implements as an SSE stream. I believe that makes sense?

docs/proposals/protobuf-grpc-migration.md
```diff
@@ -0,0 +1,218 @@
...
+plugins:
+  python_grpc:
+    type: pull
+    command: uv
+    args: [...]
+    queueTimeoutSecs: 30       # Max time in queue (default 30, null = no timeout)
+    executionTimeoutSecs: null  # Max execution time (default null = no limit)
+```
+
+**Metrics** (6 new metrics with `task.` prefix):
```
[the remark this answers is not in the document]
bjchambers: Yes. These are orchestrator metrics. Workers will report metrics they want, either generically (eg., the langflow worker) or for specific components, et.

proto/stepflow/v1/blobs.proto
```diff
@@ -0,0 +1,119 @@
...
+// The gRPC layer uses `oneof` to carry either JSON or raw bytes efficiently.
+service BlobService {
+  // Store a blob and return its content-based ID.
+  rpc PutBlob(PutBlobRequest) returns (PutBlobResponse) {
+    option (google.api.http) = {
+      post: "/blobs"
+      body: "*"
+    };
+    option (gnostic.openapi.v3.operation) = {
+      summary: "Store a blob"
```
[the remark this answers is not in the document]
bjchambers: I think if we look at the openapi documentation, the comments on the `PutBlobResponse` type also show up. I think the same is true for any generated code. So, I don't think we nede to describe that in more detail here -- it is already pulling the comments from the request/response types.

docs/proposals/protobuf-grpc-migration.md
```diff
@@ -0,0 +1,218 @@
...
+- All service implementations use structured errors
+
+### Phase 5b: Task lifecycle hardening (complete)
+
+Two-phase timeout and heartbeat system covering three failure modes:
+
+**Failure modes:**
+1. **Queuing failure** — task dispatched but no worker picks it up. Guarded by configurable queue timeout (default 30s).
+2. **Worker crash** — worker called `StartTask` but hard-crashed. Detected by heartbeat timeout (5s without a heartbeat).
+3. **Worker exception** — component raised an error caught by the worker. Reported via `CompleteTask` with error string.
```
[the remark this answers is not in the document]
bjchambers: I'm less concerned about this. Generally, orchestrators should be more stable than workers since all they do is coordinate the work. And, even if the orchestrator *does* fail, the task can be re-executed. The main purpose for this mechanism would be to avoid duplicate work, but I don't think this is necessary complexity yet.

proto/stepflow/v1/flows.proto
bjchambers: Investigated, and this seems to be *not* suggested. Simple types should use primitives for ergonomics in the generated code.

Issue #754, opened 2026-03-13
Verify docling-step-worker-proto deployment, reporting to o11y stack, and complete functionality


zznate:
A bug hunting issue for the new docling-step-worker-proto based on the gRPC/Protobuf refactor.

PR #758, opened 2026-03-13
feat: Enable gRPC pull-based worker deployment to K8s (#754)

zznate:
- Add STEPFLOW_GRPC_BIND_ADDRESS env var to grpc_server.rs so the gRPC listener can bind 0.0.0.0 for cross-pod access in K8s
- Add STEPFLOW_ORCHESTRATOR_GRPC_URL env var to grpc_plugin_config.rs so task assignments advertise the service DNS name instead of the local bind address
- Fix heartbeat race condition in pending_tasks.rs where newly started tasks were timed out before their first heartbeat by adding a grace period equal to HEARTBEAT_TIMEOUT; add regression test
- Add __main__ guard to docling_proto_step_worker/server.py for python -m invocation from Docker entrypoint
- Add multi-stage Dockerfile.stepflow-server.build for building the server binary from source inside Docker
- Update K8s orchestrator manifests with new env vars and adjust NodePort for existing Kind cluster compatibility

### DIFF
--- a/examples/production/k8s/stepflow-docling-grpc/orchestrator/deployment.yaml
+++ b/examples/production/k8s/stepflow-docling-grpc/orchestrator/deployment.yaml
@@ -128,10 +128,7 @@ spec:
         # Pull plugin gRPC bind address — must be 0.0.0.0 for external worker pods
         - name: STEPFLOW_BIND_ADDRESS
           value: "0.0.0.0:7837"
-        # Advertised URL for workers to call back (service DNS, not bind address)
-        - name: STEPFLOW_ORCHESTRATOR_URL
-          value: "docling-orchestrator.stepflow-docling-grpc.svc.cluster.local:7837"
-        # Pod metadata
+        # Pod metadata (defined first so later env vars can reference them)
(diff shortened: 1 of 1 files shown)


stepflow-rs/crates/stepflow-grpc/src/grpc_plugin_config.rs
```diff
@@ -242,8 +242,11 @@ impl stepflow_plugin::Plugin for PullPlugin {
             execution_timeout,
         );
 
-        // Set the orchestrator URL so task assignments carry the shared server address
-        inner_plugin.set_orchestrator_url(server_address.clone());
+        // Set the orchestrator URL so task assignments carry the shared server address.
+        // In K8s, workers need the service DNS name rather than the local bind address.
+        let advertised_address = std::env::var("STEPFLOW_ORCHESTRATOR_GRPC_URL")
```
bjchambers: Should just be STEPFLOW_ORCHESTRATOR_URL, and this needs to be more like `STEPFLOW_INSTANCE_URL` or something.

Issue #757, opened 2026-03-13
feat: Update observability for docling-proto-step-worker gRPC pull transport


zznate:
## Summary

The migration from HTTP JSON-RPC push to gRPC pull transport (`type: pull`) changes which metrics are available and how trace context propagates for the docling worker. The OTel Collector pipeline and docling application metrics are transport-agnostic and need no changes, but the Grafana dashboard, trace propagation verification, and service naming need attention.

## Context

PR #734 introduces 6 task lifecycle metrics on the orchestrator side (`task.queue_duration_seconds`, `task.execution_duration_seconds`, `task.queue_timeout_total`, `task.execution_timeout_total`, `task.success_total`, `task.failure_total`) that replace the HTTP server metrics the existing Grafana docling dashboard depends on. The new `stepflow-docling-grpc` K8s namespace will use these metrics exclusively — there are no HTTP server metrics because workers are gRPC clients, not HTTP servers.

### 1. Verify trace context propagation across gRPC boundary

**Priority: High** — Without this, traces in Jaeger won't show connected spans across the facade → orchestrator → worker chain.

In the HTTP push model, trace context propagated via W3C `traceparent` headers, automatically instrumented by `HTTPXClientInstrumentor`. In the pull model, trace context must flow through the `TaskAssignment` proto message (via `ExecutionContext` or `TaskContext` fields in `common.proto`) and be extracted by `grpc_worker.py` to create child spans.

Verify on the merged `protobuf-exploration` branch:
- [ ] Traces in Jaeger show facade → orchestrator → worker as connected parent-child spans
- [ ] The orchestrator's `run_id` appears in worker spans (enables deep-linking from Grafana to Jaeger)
- [ ] `grpc_worker.py` extracts trace/span IDs from `TaskAssignment` and creates properly parented spans
- [ ] gRPC metadata-based context propagation works for `StartTask`, `Heartbeat`, and `CompleteTask` calls back to the orchestrator

If trace propagation is broken or missing, the fix is in `grpc_worker.py` — extract trace context from the `TaskAssignment` and set it as the parent span before invoking component handlers.

### 2. Update Grafana docling dashboard

**Priority: Medium** — The existing `configmap-dashboard-docling.yaml` queries HTTP server metrics that don't exist in the gRPC deployment.

Current dashboard queries like:
```promql
stepflow_http_server_active_requests{exported_job="stepflow/docling-serve"}
```

Need to be replaced with the new task lifecycle metrics:

[...]

PR #774, opened 2026-03-16
feat: Add observability for gRPC pull transport (#757)

zznate:
- Add PullTaskQueue operational metrics on the Rust orchestrator side: task.queue_depth, task.connected_workers, task.dispatched_total
- Extract parent SpanContext from orchestrator observability fields and wrap task execution in a child span for distributed tracing continuity
- Add worker-side OTel metrics (tasks_pulled, tasks_completed, heartbeats_sent, connection_status) with lazy init to avoid the Python OTel SDK proxy-meter no-op issue
- Add Grafana dashboard for gRPC pull transport with 7 rows: Chunking, Document Processing, Task Lifecycle, Queue Health, Step Retries, Blob Storage, Traces & Logs
- Update K8s CLAUDE.md to focus on stepflow-docling-grpc architecture, image build instructions, and gRPC pull transport metrics reference

### DIFF
--- a/sdks/python/stepflow-py/src/stepflow_py/config/_generated_config.py
+++ b/sdks/python/stepflow-py/src/stepflow_py/config/_generated_config.py
@@ -148,7 +148,7 @@ class PullPluginConfig(Struct, kw_only=True, tag_field='type', tag='pull'):
         Annotated[
             str | None,
             Meta(
-                description="Queue name the worker uses to receive tasks (matches plugin key in orchestrator config).\nDefaults to the plugin's key name in the config."
+                description="Queue name the worker uses to receive tasks. Required — must match\n`STEPFLOW_QUEUE_NAME` in the worker's environment."
             ),
         ]
         | UnsetType


stepflow-rs/crates/stepflow-observability/src/metrics.rs
```diff
@@ -302,3 +302,56 @@ pub fn record_task_success(component: &str) {
...
+static TASK_QUEUE_DEPTH: LazyLock<UpDownCounter<i64>> = LazyLock::new(|| {
+    let meter = global::meter("stepflow");
+    meter
+        .i64_up_down_counter("task.queue_depth")
+        .with_description("Current number of tasks waiting in the pull queue")
+        .build()
+});
+
+/// Number of workers currently connected to the pull queue.
+static TASK_CONNECTED_WORKERS: LazyLock<UpDownCounter<i64>> = LazyLock::new(|| {
```
bjchambers: I'm not sure we can reliably get this, and even if we can, it is a *queue* property which won't remain once we have NATS.

Issue #849, opened 2026-03-25
bug:  SQLite storage backend: "failed to get step statuses: Internal state store error" on read


zznate:
When the stepflow-server is configured with `storageConfig: { type: sqlite }`, writes succeed (runs are created, blobs stored, step_statuses populated), but reads of run details fail with HTTP 500:

```json
{
  "error": {
    "code": 500,
    "message": "failed to get step statuses: Internal state store error",
    "status": "INTERNAL"
  }
}
```

This affects `GET /api/v1/runs/{run_id}` and the facade's `/v1/status/poll/{task_id}` endpoint. The server starts and accepts submissions without error — the failure only manifests on read-back.

## Environment

- **stepflow-server version:** 0.12.0 (build features: `[]`)
- **Binary target:** `aarch64-unknown-linux-musl` (Alpine 3.19)
- **Rust compiler:** rustc 1.94.0 (2026-03-02)
- **Platform:** Kind cluster on macOS (Apple Silicon)
- **Config:**
  ```yaml
  storageConfig:
    type: sqlite
    databaseUrl: "sqlite:/data/stepflow.db?mode=rwc"
  ```
- **Volume:** emptyDir mounted at `/data` on the stepflow-server container

## Steps to reproduce

1. Deploy the orchestrator with sqlite storage config (see above)
2. Verify the DB is created: `kubectl exec <pod> -c stepflow-server -- ls -la /data/`
3. Submit documents via the facade (e.g., run the load test)
4. Observe that submissions are accepted (202) and runs are created
5. Poll for status — all poll requests return HTTP 500

## What works

- Server startup (SQLite DB created, tables migrated)
- Flow registration via `POST /api/v1/flows` (blob stored correctly)
- Run creation via `POST /api/v1/runs`
- Task execution (runs complete successfully at the orchestrator level)
- `GET /api/v1/runs` (list endpoint) returns run summaries correctly
- Direct SQLite queries against the DB file work fine

## What fails

- `GET /api/v1/runs/{run_id}` — returns 500
- Any endpoint that calls `get step statuses` internally

## Database analysis

The SQLite DB was copied off the pod and inspected locally. All tables exist and contain correct data:

```
Tables: _stepflow_migrations, blobs, checkpoints, journal_entries,
        run_items, runs, step_info, step_results, step_statuses

Counts: runs=11, blobs=1, step_statuses=33, journal_entries=130
```

[...]

PR #852, opened 2026-03-25
fix: Resolve SQLite storage backend read failures (#849)

zznate:
SQLite PRAGMAs (journal_mode=WAL, busy_timeout=5000) were set via one-shot queries on a single pool connection. Other pool connections retained busy_timeout=0, causing immediate SQLITE_BUSY failures on read contention. Moved PRAGMAs to SqliteConnectOptions so they apply to every connection created by the pool.

Additional changes:
- Replace row.get() with try_get() to surface decode errors instead of panicking
- Add error context (run_id, step_id, column name) throughout the step_statuses read path
- Log full error-stack chain in gRPC handlers before converting to status codes
- Add 6 round-trip tests for step_statuses (including 2MB payload and concurrent file-backed access)
- Enable SQLite persistent storage in K8s orchestrator deployment (emptyDir volume at /data, storageConfig type=sqlite)


stepflow-rs/crates/stepflow-grpc/src/services/runs_service.rs
```diff
@@ -395,7 +398,10 @@ impl RunsService for RunsServiceImpl {
             .metadata_store()
             .get_step_statuses(run_id, item_index, None)
             .await
-            .map_err(|e| grpc_err::internal(format!("failed to get step statuses: {e}")))?;
+            .map_err(|e| {
+                log::error!("get_step_statuses failed for run {run_id}: {e:?}");
```
bjchambers: Ideally we log the trace back for all exceptions in one place so we shouldn't need to add each of these.

PR #902, opened 2026-04-10
fix: Handle empty path in gRPC worker component discovery

ogabrielluiz:
## Summary

When Python SDK workers report components via `ListComponentsResult`, the `component_id` field includes a leading slash (e.g., `"/udf"`) and the `path` field is an empty string. `parse_discovery_result` treated the empty path as a valid value, so the trie builder produced double-slash paths (`"//udf"`) that never match during routing.

- Filter empty `path` strings so they fall back to the `component_id` value

### DIFF
--- a/stepflow-rs/crates/stepflow-grpc/src/queue_plugin.rs
+++ b/stepflow-rs/crates/stepflow-grpc/src/queue_plugin.rs
@@ -337,7 +337,6 @@ fn parse_discovery_result(value: &ValueRef) -> Result<Vec<ComponentInfo>> {
     for c in components_json {
         let name = c
             .get("component_id")
-            .or_else(|| c.get("name"))
             .and_then(|v| v.as_str())
             .unwrap_or_default();
         let path = c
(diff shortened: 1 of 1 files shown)


stepflow-rs/crates/stepflow-grpc/src/queue_plugin.rs
```diff
@@ -337,11 +337,13 @@ fn parse_discovery_result(value: &ValueRef) -> Result<Vec<ComponentInfo>> {
     for c in components_json {
         let name = c
             .get("component_id")
+            .or_else(|| c.get("name"))
```
bjchambers: I don't think there is a "name" field in the component info proto, so this shouldn't match anything.

ogabrielluiz: good catch, removed it. the actual fix is just the empty path filter below.

bjchambers:
You likely need to sign the ICLA to unblock the PR Checks. See the instructions https://github.com/stepflow-ai/stepflow/blob/main/CONTRIBUTING.md#getting-started.

I suspect the Langflow Integration Checks are failing because it doesn't have access to he OPENAI_API_KEY environment variable for "non-repository member" runs of the flow. I'll double check once everything else passes. But this looks good.

PR #905, opened 2026-05-11
fix: Fix langflow integration local run and component execution issues

jordanrfrazier:
- Fix plugin type from "stepflow" to "grpc" in _run_local config
- Add required queueName field to gRPC plugin config
- Fix _display_run_result to use correct field names (run_id, item_index, output)
- Fix _apply_component_input_defaults to allow empty string defaults, preventing None being passed to fields like context_id that require a non-None value
- Add STEPFLOW_MAX_RETRIES env var support to standalone server
- Restrict requires-python to <3.14 to fix uv resolution on Python 3.14
- Update uv.lock for langflow>=1.9.0 and lfx>=0.4


integrations/langflow/src/stepflow_langflow_integration/executor/custom_code_executor.py
```diff
@@ -415,7 +415,6 @@ def _apply_component_input_defaults(
                     if (
                         param_name not in component_parameters
                         and default_value is not None
-                        and default_value != ""
```
jordanrfrazier, line 418: This was for the `context_id` being empty in a request

integrations/langflow/pyproject.toml
```diff
@@ -10,7 +10,7 @@ authors = [
 maintainers = [
     {name = "Stepflow Contributors"}
 ]
-requires-python = ">=3.11"
+requires-python = ">=3.11,<3.14"
```
jordanrfrazier, line 13: Langflow doesn't support 3.14 yet

bjchambers (approved):

